PR #2, opened 2015-05-21
Cleanup

bakura10:
Hi!

This PR only contains some very minor cleanup to match ZF codestyle and small PHPDoc fixes.

### DIFF
--- a/src/AbstractSerializer.php
+++ b/src/AbstractSerializer.php
@@ -38,7 +38,7 @@ abstract class AbstractSerializer
     {
         $line    = '';
         $crFound = false;
-        while (! $stream->eof()) {
+        while (!$stream->eof()) {
             $char = $stream->read(1);
 
             if ($crFound && $char === self::LF) {
@@ -52,7 +52,7 @@ abstract class AbstractSerializer
             }
 
             // LF in isolation
-            if (! $crFound && $char === self::LF) {
+            if (!$crFound && $char === self::LF) {
                 throw new UnexpectedValueException('Unexpected line feed detected');
             }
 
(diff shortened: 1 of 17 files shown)


src/AbstractSerializer.php
```diff
@@ -38,7 +38,7 @@ protected static function getLine(StreamInterface $stream)
     {
         $line    = '';
         $crFound = false;
-        while (! $stream->eof()) {
+        while (!$stream->eof()) {
```
weierophinney: Honestly, this is a personal preference thing, and I've actually been changing these in the ZF2 code base to `! $...` when I come across them. Let's leave these as-is, please.

src/Request.php
```diff
@@ -40,9 +40,7 @@ public function __construct($uri = null, $method = null, $body = 'php://memory',
     public function getHeaders()
     {
         $headers = $this->headers;
-        if (! $this->hasHeader('host')
-            && ($this->uri && $this->uri->getHost())
-        ) {
+        if (!$this->hasHeader('host') && ($this->uri && $this->uri->getHost())) {
```
weierophinney: This was easier to read as two lines.

src/Request.php
```diff
@@ -54,10 +52,8 @@ public function getHeaders()
      */
     public function getHeader($header)
     {
-        if (! $this->hasHeader($header)) {
-            if (strtolower($header) === 'host'
-                && ($this->uri && $this->uri->getHost())
-            ) {
+        if (!$this->hasHeader($header)) {
+            if (strtolower($header) === 'host' && ($this->uri && $this->uri->getHost())) {
```
weierophinney: This was easier to read as two lines.

src/ServerRequestFactory.php
```diff
@@ -227,7 +226,6 @@ public static function marshalHeaders(array $server)
      * @param array $server
      * @param MessageInterface $request
      * @return Uri
-     * @deprecated as of 0.7.0; use marshalUriFromServer() instead.
```
weierophinney: We should remove the method, really.

Ocramius: 2.0.0 :-\

src/UploadedFile.php
```diff
@@ -60,40 +60,42 @@ public function __construct($streamOrFile, $size, $errorStatus, $clientFilename
...
             throw new InvalidArgumentException('Invalid size provided for UploadedFile; must be an int');
         }
+
         $this->size = $size;
 
-        if (! is_int($errorStatus)
-            || 0 > $errorStatus
-            || 8 < $errorStatus
-        ) {
+        if (!is_int($errorStatus) || 0 > $errorStatus || 8 < $errorStatus) {
```
weierophinney: I prefer multiple lines here, as it's far simpler to see and understand the specific conditions than when on the same line.

src/Uri.php
```diff
@@ -275,7 +275,7 @@ public function withHost($host)
      */
     public function withPort($port)
     {
-        if (! (is_integer($port) || (is_string($port) && is_numeric($port)))) {
+        if (!is_numeric($port)) {
```
weierophinney: I could have sworn `is_numeric()` did not return `true` for non-string values, but I just tested, and it does. Good change.

src/ServerRequestFactory.php
weierophinney: I should have removed it before creating the 1.0.0 tag. I messed up. :-(

bakura10: Okay, I’m too lazy to revert all the space changes, maybe it’s easier for you to pick the two-three interesting changes?

Too bad to have deprecated methods in 1.0.0… :( BUT you can tag 2.0.0 as that’s the main advantage of splitting repositories (or let’s remove them and add a BC notice… this thing has been released yesterday, I hope nobody already use that in production and using deprecated features)

weierophinney: Sure, I'll cherry-pick when I review for merge.

src/ServerRequestFactory.php
```diff
@@ -464,9 +468,8 @@ private static function createUploadedFileFromSpec(array $value)
      * @param array $files
      * @return UploadedFileInterface[]
      */
-    private static function normalizeNestedFileSpec(array $files)
+    private static function normalizeNestedFileSpec(array $files = [])
     {
-        $files = [];
```
weierophinney: Actually, I think you've spotted an error; the value shouldn't be reset at the start of the method!

PR #4, opened 2015-05-21
Fix return annotation

franzliedke:
That method does not return void.


franzliedke:
Whoops, apparently this is already part of #2.

PR #5, opened 2015-05-21
added more tests

samsonasik:


test/ServerRequestTest.php
```diff
@@ -97,7 +97,19 @@ public function testRemovingAttributeReturnsCloneWithoutAttribute($request)
         $this->assertNull($new->getAttribute('foo', null));
     }
 
-    public function testUsesProvidedConstructorArguments()
+    public function provideMethods()
+    {
+        return array(
```
bakura10: Use short array notation (Diactoros is PHP 5.5+)

weierophinney: Agreed; that's one thing I've enjoyed about developing on phly/http and now diactoros. :)

samsonasik: done ;)

Warm regards,

Abdul Malik Ikhsan

Pada 22 Mei 2015, pukul 21.52, weierophinney <notifications@github.com> menulis:

Maks3w: Please avoid review comments about styling. If the CS tool not yield there is nothing to discuss.

PR #9, opened 2015-05-22
Fix error when accessing a single attribute before any have been added.

dolfelt:
When using `getAttribute` before `withAttribute` has been called the following error happens:

`array_key_exists() expects parameter 2 to be array, null given`

This fix makes sure the array is initialized in the `ServerRequest` class. I'm open to other suggestions for solving this issue.

cc: @weierophinney

### DIFF
--- a/src/ServerRequest.php
+++ b/src/ServerRequest.php
@@ -35,7 +35,7 @@ class ServerRequest implements ServerRequestInterface
     /**
      * @var array
      */
-    private $attributes;
+    private $attributes = [];
 
     /**
      * @var array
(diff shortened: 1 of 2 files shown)


Ocramius:
:+1: thanks!

`master`: zendframework/zend-diactoros@15d8b842315b3bf5fee6f97f81a8ce9abc6852e0
`develop`: zendframework/zend-diactoros@dd3ec9c3abe0f8ce444ee751eefd0655a4b03680

dolfelt:
Awesome! Nice, quick turn around.

PR #10, opened 2015-05-22
[WIP] Feature: "relative" stream

mtymek:
Initial implementation of "relative" stream, representing subpart of another stream, to be reviewed.

It is a first step towards parsing large bodies that may not fit in PHP's available memory. 

There are at least two concerns for now:

* using relative stream will change original stream's internal pointer. This could be avoided by adding calls to `tell()` and `seek()`, but it would also complicate things a lot (need to check if stream is seekable, etc). Should I care it at all?
* I don't like the name - can someone come up with something better than "RelativeStream" :) ?

Note that test suite is not finished - I will do it during the weekend.

### DIFF
--- a/src/AbstractSerializer.php
+++ b/src/AbstractSerializer.php
@@ -114,9 +114,8 @@ abstract class AbstractSerializer
             $headers[$currentHeader][] = $value . ltrim($line);
         }
 
-        $body = new RelativeStream($stream, $stream->tell());
-
-        return [$headers, $body];
+        // use RelativeStream to avoid copying initial stream into memory
+        return [$headers, new RelativeStream($stream, $stream->tell())];
     }
 
     /**
--- a/src/AbstractSerializer.php
+++ b/src/AbstractSerializer.php
@@ -114,13 +114,7 @@ abstract class AbstractSerializer
             $headers[$currentHeader][] = $value . ltrim($line);
         }
 
-        $body = new Stream('php://temp', 'wb+');
-        if (! $stream->eof()) {
-            while ($data = $stream->read(4096)) {
-                $body->write($data);
-            }
-            $body->rewind();
-        }
+        $body = new RelativeStream($stream, $stream->tell());
 
         return [$headers, $body];
     }
--- a/src/RelativeStream.php
+++ b/src/RelativeStream.php
@@ -1,0 +1,165 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @see       http://github.com/zendframework/zend-diactoros for the canonical source repository
+ * @copyright Copyright (c) 2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-diactoros/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Diactoros;
+
+use Psr\Http\Message\StreamInterface;
+
+class RelativeStream implements StreamInterface
+{
+    /**
+     * @var StreamInterface
+     */
+    private $decodatedStream;
+
+    /**
+     * @var int
+     */
+    private $offset;
+
+    /**
+     * Class constructor
+     *
+     * @param StreamInterface $decodatedStream
+     * @param $offset
+     */
+    public function __construct(StreamInterface $decodatedStream, $offset)
+    {
+        $this->decodatedStream = $decodatedStream;
+        $this->offset = $offset;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function __toString()
+    {
+        $this->seek(0);
+        return $this->getContents();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function close()
+    {
+        $this->decodatedStream->close();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function detach()
+    {
+        return $this->decodatedStream->detach();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getSize()
+    {
+        return $this->decodatedStream->getSize() - $this->offset;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function tell()
+    {
+        return $this->decodatedStream->tell() - $this->offset;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function eof()
+    {
+        return $this->decodatedStream->eof();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function isSeekable()
+    {
+        return $this->decodatedStream->isSeekable();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function seek($offset, $whence = SEEK_SET)
+    {
+        switch ($whence) {
+            case SEEK_SET:
+                $basePos = $this->offset;
+                break;
+            default:
+                $basePos = 0;
+        }
+        return $this->decodatedStream->seek($offset + $basePos, $whence);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function rewind()
+    {
+        return $this->seek(0);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function isWritable()
+    {
+        return $this->decodatedStream->isWritable();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function write($string)
+    {
+        return $this->decodatedStream->write($string);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function isReadable()
+    {
+        return $this->decodatedStream->isReadable();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function read($length)
+    {
+        return $this->decodatedStream->read($length);
+    }
+
+    /**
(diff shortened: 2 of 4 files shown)


src/RelativeStream.php
```diff
@@ -0,0 +1,165 @@
...
+     * @param $offset
+     */
+    public function __construct(StreamInterface $decodatedStream, $offset)
+    {
+        $this->decoratedStream = $decodatedStream;
+        $this->offset = $offset;
+    }
+
+    /**
+     * @inheritdoc
```
samsonasik: {@inheritdoc} with {}

mtymek:
@samsonasik thanks for spotting - updated.

src/RelativeStream.php
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * @var int
+     */
+    private $offset;
+
+    /**
+     * Class constructor
+     *
+     * @param StreamInterface $decodatedStream
+     * @param $offset
```
samsonasik: @param paramType $offset

src/RelativeStream.php
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * Class constructor
+     *
+     * @param StreamInterface $decodatedStream
+     * @param int $offset
+     */
+    public function __construct(StreamInterface $decodatedStream, $offset)
+    {
+        $this->decoratedStream = $decodatedStream;
+        $this->offset = $offset;
```
Ocramius: cast to `int` or check that it is an unsigned integer

src/RelativeStream.php
```diff
@@ -0,0 +1,165 @@
...
+    {
+        return $this->decoratedStream->isSeekable();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function seek($offset, $whence = SEEK_SET)
+    {
+        switch ($whence) {
```
Ocramius: Avoid using the `switch` here. An `if` and early return is sufficient and simpler to read.

src/RelativeStream.php
```diff
@@ -0,0 +1,165 @@
...
+ * @see       http://github.com/zendframework/zend-diactoros for the canonical source repository
+ * @copyright Copyright (c) 2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-diactoros/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Diactoros;
+
+use Psr\Http\Message\StreamInterface;
+
+class RelativeStream implements StreamInterface
```
Ocramius: Please accurately document this class and its use-case, perhaps even with an `@example`

src/AbstractSerializer.php
```diff
@@ -114,13 +114,7 @@ protected static function splitStream(StreamInterface $stream)
...
         }
 
-        $body = new Stream('php://temp', 'wb+');
-        if (! $stream->eof()) {
-            while ($data = $stream->read(4096)) {
-                $body->write($data);
-            }
-            $body->rewind();
-        }
+        $body = new RelativeStream($stream, $stream->tell());
```
Ocramius: Return directly, no need for an assignment.

Ocramius: Also document (comment) why you are doing this (avoiding reading from the initial stream to copy data into memory)

test/RelativeStreamTest.php
```diff
@@ -0,0 +1,144 @@
...
+ * @license   https://github.com/zendframework/zend-diactoros/blob/master/LICENSE.md New BSD License
+ */
+
+namespace ZendTest\Diactoros;
+
+use PHPUnit_Framework_TestCase as TestCase;
+use Zend\Diactoros\RelativeStream;
+use Zend\Diactoros\Stream;
+
+class RelativeStreamTest extends TestCase
```
Ocramius: Please add an `@covers` docblock to prevent coverage leakage

src/RelativeStream.php
```diff
@@ -0,0 +1,168 @@
...
+     */
+    private $offset;
+
+    /**
+     * Class constructor
+     *
+     * @param StreamInterface $decoratedStream
+     * @param int $offset
+     */
+    public function __construct(StreamInterface $decoratedStream, $offset)
```
Ocramius: Would it make sense to also provide `$limit = null`? Or maybe am I being a feature creep here? /cc @weierophinney

src/AbstractSerializer.php
```diff
@@ -114,15 +114,8 @@ protected static function splitStream(StreamInterface $stream)
...
-        if (! $stream->eof()) {
-            while ($data = $stream->read(4096)) {
-                $body->write($data);
-            }
-            $body->rewind();
-        }
-
-        return [$headers, $body];
+        // use RelativeStream to avoid copying initial stream into memory
+        return [$headers, new RelativeStream($stream, $stream->tell())];
```
Ocramius: This change has no matching test case. Yes, it is an implementation detail, but we should also verify that `$stream->expects($this->never())->method('read')` in order to validate the scenario.

src/RelativeStream.php
mtymek: I would leave it until it is needed - can be added any time.

Ocramius: makes sense. We can build separate stream classes for that use-case if it ever comes up.

src/AbstractSerializer.php
mtymek: `read()` is actually used for scanning the header. What about testing if instance of `RelativeStream` is returned?

Ocramius: @mtymek that's a separate issue. What you are actually verifying here is that you're not sequentially scanning through a huge file output, but are instead lazily evaluating it.

Asserting that a `RelativeStream` is returned gives you no such guarantee.

Ocramius: What you can do is checking that the body isn't read

Ocramius: One way to test this is creating a long string body and verifying that `read()` was not called more than N times (thinking aloud)

mtymek: This could work. Headers are scanned at one character at the time (`read(1)`), so number of calls to `read` should equal to headers size.

mtymek:
Added test.

Ocramius:
Looking good. Can anyone else proof-read it before I merge?

gianarb:
:+1: IMHO

mtymek:
Pushed CS fix to prevent build from failing.

test/RelativeStreamTest.php
```diff
@@ -0,0 +1,147 @@
...
+use Zend\Diactoros\Stream;
+
+/**
+ * @covers \Zend\Diactoros\RelativeStream
+ */
+class RelativeStreamTest extends TestCase
+{
+    public function testToString()
+    {
+        $decorated = $this->prophesize(Stream::class);
```
weierophinney: This is the first use I've seen of Prophecy. Guess I have a bit of studying to do in the near future!

weierophinney:
:+1:

Ocramius:
Merged, thanks!

`master`: zendframework/zend-diactoros@7f580a10e6b11d79126c6344cc25f98a93574bd4
`develop`: zendframework/zend-diactoros@c5d11a697fe11f36aa61fa8fbd112df656bb864e

test/RelativeStreamTest.php
mtymek: Once you start using it, you will never want to go back :-) Start here: https://thephp.cc/news/2015/02/phpunit-4-5-and-prophecy

Ocramius: I'll actually say that I don't like neither mockery nor prophecy mainly due to the lack of autocompletion. Yes, feel free to throw stuff at me now :-P

src/RelativeStream.php
```diff
@@ -0,0 +1,168 @@
...
+use Psr\Http\Message\StreamInterface;
+
+/**
+ * Class RelativeStream
+ *
+ * Wrapper for default Stream class, representing subpart (starting from given offset) of initial stream.
+ * It can be used to avoid copying full stream, conserving memory.
+ * @example see Zend\Diactoros\AbstractSerializer::splitStream()
+ */
+class RelativeStream implements StreamInterface
```
Ocramius: @weierophinney what do you think about `final` here? No methods defined besides the interface.

test/RelativeStreamTest.php
mtymek: Sometimes I "cheat" in my tests to get autocompletion:

```php
/** @var ObjectProphecy|Stream $decorated */
        $decorated = $this->prophesize(Stream::class);
        $decorated->seek(100, SEEK_SET)->shouldBeCalled();
        $decorated->getContents()->shouldBeCalled()->willReturn('foobarbaz');
```

weierophinney: I hate those comments. :)

src/RelativeStream.php
weierophinney: I think marking it final makes sense.

Ocramius: Did so in #20

PR #14, opened 2015-05-24
Use php://temp instead of php://memory by default in Response

dunglas:
Avoid problem of exceeded memory when dealing with large responses.

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -29,7 +29,7 @@ class Request implements RequestInterface
      * @param array $headers Headers for the message, if any.
      * @throws \InvalidArgumentException for any invalid value.
      */
-    public function __construct($uri = null, $method = null, $body = 'php://memory', array $headers = [])
+    public function __construct($uri = null, $method = null, $body = 'php://temp', array $headers = [])
     {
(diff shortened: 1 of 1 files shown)


Ocramius:
@dunglas pardon the ignorance, but will `php://temp` write to memory until a limit is reached? I just want to avoid file I/O on each usage, and couldn't find clear documentation about this wrapper ;-)

dunglas:
@Ocramius yes. The default limit is 2 Mo (can be configured by adding using a syntax like `php://temp:NN` where NN is the limit in bytes).

The doc is available here: http://php.net/manual/en/wrappers.php.php I struggled to find it too.

Ocramius:
Thanks!

Ocramius:
Merged, thanks!

`master`: zendframework/zend-diactoros@f347705d8debd27ba9855517d79537f0d030737b
`develop`: zendframework/zend-diactoros@0d5bc9914654f136fb8108ad1c160e9d9714128e

sasezaki:
Title is `Response`, but changed is `Request` ?

dunglas:
@sasezaki good catch my title is wrong. I update it.

PR #15, opened 2015-05-24
Raise exception on write() if stream is not writeable

### LINKED ISSUE #13: Raise exception on write() if stream is not writeable
As noted in [a comment on zfr-oauth2-server](https://github.com/zf-fr/zfr-oauth2-server/pull/31/files#r30953668), `Stream::write()` does not currently raise an exception when writing to a stream that is not writeable. This causes silent errors when working with the `php://` streams, as they will not raise notices for operations they cannot fulfill.

mtymek:
Fixes #13.

I had to change how `isWritable` does the check, as `is_writable($meta['uri'])` is unreliable for TEMP and MEMORY streams.

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -224,7 +224,13 @@ class Stream implements StreamInterface
         $meta = stream_get_meta_data($this->resource);
         $mode = $meta['mode'];
 
-        return (strstr($mode, 'w') || strstr($mode, '+'));
+        return (
+            strstr($mode, 'x')
+            || strstr($mode, 'w')
+            || strstr($mode, 'c')
+            || strstr($mode, 'a')
+            || strstr($mode, '+')
+        );
     }
 
     /**
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -297,7 +297,7 @@ class Stream implements StreamInterface
     public function getContents()
     {
         if (! $this->isReadable()) {
-            return '';
+            throw new RuntimeException('Stream is not readable');
         }
 
         $result = stream_get_contents($this->resource);
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -222,7 +222,9 @@ class Stream implements StreamInterface
         }
 
         $meta = stream_get_meta_data($this->resource);
-        return is_writable($meta['uri']);
+        $mode = $meta['mode'];
+
+        return (strstr($mode, 'w') || strstr($mode, '+'));
     }
 
     /**
@@ -232,6 +234,10 @@ class Stream implements StreamInterface
     {
         if (! $this->resource) {
             throw new RuntimeException('No resource available; cannot write');
+        }
+
+        if (! $this->isWritable()) {
+            throw new RuntimeException('Stream is not writable');
         }
 
         $result = fwrite($this->resource, $string);
--- a/test/StreamTest.php
+++ b/test/StreamTest.php
@@ -246,6 +246,12 @@ class StreamTest extends TestCase
         $this->assertFalse($stream->isWritable());
     }
 
+    public function testIsWritableReturnsTrueForWritableMemoryStream()
+    {
+        $stream = new Stream("php://temp", "r+b");
+        $this->assertTrue($stream->isWritable());
+    }
+
     public function testWriteRaisesExceptionWhenStreamIsDetached()
     {
         $this->tmpnam = tempnam(sys_get_temp_dir(), 'diac');
@@ -254,6 +260,13 @@ class StreamTest extends TestCase
         $stream = new Stream($resource);
         $stream->detach();
         $this->setExpectedException('RuntimeException', 'No resource');
+        $stream->write('bar');
+    }
+
+    public function testWriteRaisesExceptionWhenStreamIsNotWritable()
+    {
+        $stream = new Stream('php://memory', 'r');
+        $this->setExpectedException('RuntimeException', 'Stream is not writable');
         $stream->write('bar');
     }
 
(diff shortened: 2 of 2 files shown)


src/Stream.php
```diff
@@ -222,7 +222,9 @@ public function isWritable()
         }
 
         $meta = stream_get_meta_data($this->resource);
-        return is_writable($meta['uri']);
+        $mode = $meta['mode'];
+
+        return (strstr($mode, 'w') || strstr($mode, '+'));
```
Ocramius: What are *all* the possible valid `$mode` strings? Can `w` be before `+`?

mtymek: Answer is complex, depending on your definition  of "valid". If by "valid" you mean everything that PHP accepts, then:
* for TEMP and MEMORY streams, `mode` is normalized, you always get `rb` or `w+b`
* for files PHP will accept even things like this: `$f2 = fopen('/tmp/test', 'ah()+');`, and report them under `mode`

So, apparently in order to satisfy `fopen` docs, we have to check if any of characters `+, a, c, w, x` is present in `mode`. If you agree, I can post fix in a moment.

Yes, `w` can be before `+`. Not sure why you ask though.

src/Stream.php
```diff
@@ -60,7 +60,7 @@ public function __construct($stream, $mode = 'r')
     public function __toString()
     {
         if (! $this->isReadable()) {
-            return '';
+            throw new RuntimeException('Stream is not readable');
```
weierophinney: Actually, we cannot throw exceptions here; see [the PHP manual](http://php.net/manual/en/language.oop5.magic.php#object.tostring). The correct behavior here is to return an empty string.

test/StreamTest.php
```diff
@@ -75,13 +75,14 @@ public function testPassingInvalidStreamResourceToConstructorRaisesException()
         $stream = new Stream(['  THIS WILL NOT WORK  ']);
     }
 
-    public function testStringSerializationReturnsEmptyStringWhenStreamIsNotReadable()
+    public function testStringSerializationRisesExceptionWhenStreamIsNotReadable()
```
weierophinney: Revert this test per the previous comment.

test/StreamTest.php
```diff
@@ -290,13 +304,14 @@ public function testReadReturnsEmptyStringWhenAtEndOfFile()
...
-    public function testGetContentsReturnsEmptyStringIfStreamIsNotReadable()
+    public function testGetContentsRisesExceptionIfStreamIsNotReadable()
     {
+        $this->setExpectedException('RuntimeException');
         $this->tmpnam = tempnam(sys_get_temp_dir(), 'diac');
         file_put_contents($this->tmpnam, 'FOO BAR');
         $resource = fopen($this->tmpnam, 'w');
         $stream = new Stream($resource);
-        $this->assertEquals('', $stream->getContents());
+        $stream->getContents();
```
Ocramius: BC break?

weierophinney: Yes and no. I can see an argument that the current behavior is broken, as it leads to unexpected results.

BTW, @mtymek - move the `setExpectedException()` call to immediately before the operation that will raise the exception; otherwise, an exception happening earlier will lead to a false positive result.

src/Stream.php
mtymek: Uhm, that's what I asked about in #16 :)

weierophinney: I know; it was seeing it in the context of the method that made me remember why I wasn't throwing an exception previously here. Since the exception is non-recoverable, we must cast to an empty string instead.

mtymek: Ah, now I remember the horror of debugging ZF1's view helper that throwing exception...

mtymek:
Updated!

src/Stream.php
```diff
@@ -222,7 +222,15 @@ public function isWritable()
...
 
         $meta = stream_get_meta_data($this->resource);
-        return is_writable($meta['uri']);
+        $mode = $meta['mode'];
+
+        return (
+            strstr($mode, 'x')
+            || strstr($mode, 'w')
+            || strstr($mode, 'c')
+            || strstr($mode, 'a')
```
Ocramius: All these modes need to be checked

mtymek:
Added tests for all modes that make sense; did the same for `isReadable`.

Ocramius:
Thanks @mtymek

test/StreamTest.php
```diff
@@ -246,6 +246,111 @@ public function testIsWritableReturnsFalseWhenStreamIsDetached()
...
+            ['x+',  false, true],
+            ['x+b', false, true],
+            ['xb',  false, true],
+        ];
+    }
+
+    private function findNonExistentTempName()
+    {
+        while (true) {
+            $tmpnam = sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'diac' . uniqid();
```
Ocramius: There is `tempnam` for this

mtymek: See docs - `tempnam` creates new file. If you open it with `x` mode, you get a warning:

```bash
PHP Warning:  fopen(/tmp/zrazAZD39N): failed to open stream: File exists in phar:///home/mat/bin/boris.phar/lib/Boris/EvalWorker.php(133) : eval()'d code on line 1
```

Ocramius: good point, didn't know :)

Ocramius:
@mtymek merged, thanks!

`master`: zendframework/zend-diactoros@d7d77518cc86ac509747a11a5953cd3668d68a5a
`develop`: zendframework/zend-diactoros@4718cf47d688f8c1a9091bacc433c52451abcb49

test/StreamTest.php
mtymek: Me neither, I realized this when my tests started to fail :-)

PR #25, opened 2015-05-26
Implemented bookdown.io docs

weierophinney:
This patch separates the various sections of README.md into discrete markdown files, and introduces [bookdown](https://github.com/bookdown/Bookdown.Bookdown) as a development dependency for generating an HTML manual. **Update**: `bookdown/bookdown` was removed as a dependency; we will recommend installing it globally if you wish to render documentation.

I have introduced custom templates and a custom Bootstrap-based CSS assets to provide a cleanly rendered manual for consumption. **UPDATE**: These were removed.

Questions I have:

- Should the templates and CSS be included? These can be omitted, and the content rendered cleanly, albeit very plainly, without them, and we could have those as part of a master "manual" project instead.
  **UPDATE**: the master templates and CSS will be in the [documentation](https://github.com/zendframework/documentation) repository.
- Does the structure seem reasonable? does it seem reasonable if we decide not to ship templates and CSS?
  **UPDATE**: The only change to the structure since the original commit for this PR was to move the `bookdown.json` up a level.


weierophinney:
Some screenshots:

usage - google chrome_009

and fully responsive:
usage - google chrome_010

weierophinney:
Ping @ezimuel , @Maks3w and @GeeH , as I think you'd each be interested.

The key benefit Bookdown provides is that we can have docs for each component *in* the component repo, but still have a master "manual" repo (or part of the website) that compiles them into a single larger manual. Additionally, it's easily templatable, which means we can embed it in the site, rather than having to do the crazy DOM-munging we currently do.

Thoughts?

GeeH:
:+1: :+1: :+1: :+1:

GeeH:
I'll definitely be looking at this as soon as I can.

weierophinney:
Another thought: we could likely have Travis-CI push compiled docs or trigger a build on success, too. :)

manuakasam:
Looks good. On a related side-note: I'd highly suggest a no-feature-without-documentation for core-libraries like these! Let's do it right from the beginning this time :)

weierophinney:
@manuakasam that's exactly what we want to enable, and part of what's prompting this PR. :)

gianarb:
Are you sure about this choice?
ReadTheDocs support markdown and custom theme.. :) maybe we can use it..

weierophinney:
@gianarb Yes, pretty sure.

- rst doesn't render well on GitHub, particularly if you rely on extensions. This makes casual browsing and editing difficult.
- rst does not give us an easy path to combine docs from multiple repos, which is a huge concern after splittng them.
- multiple locations to view docs has led to uncertainty about the canonical source for rendered docs.
- rst requires additional tools and languages in order to compile documentation, making it harder for contributors to evaluate their own changes.


In my experience, the move to rst was a good one coming off of docbook. I think moving to markdown will be even better, particularly considering its ubiquity on developer sites and tooling.

gianarb:
`ReadTheDocs support markdown`
https://github.com/gianarb/lean-queue/tree/master/docs :) 
In this example my docs in in md :)

https://docs.readthedocs.org/en/latest/index.html

doc/templates/head.php
```diff
@@ -0,0 +1,7 @@
+<head>
+    <title><?php echo $this->page->getTitle(); ?></title>
+    <link rel="stylesheet" href="assets/css/bootstrap.min.css" />
+    <link rel="stylesheet" href="assets/css/bootstrap-theme.min.css" />
+    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/8.6/styles/default.min.css">
```
Maks3w: If external resources loading is expected then bootstrap could be loaded from CDN too and only host the customizations

weierophinney: I created a custom bootstrap theme that omitted most features and used or
color scheme, but yes, could definitely do that.

Maks3w:
I have to review this many times until to agree/disagree.

By the way I'll write a list of things as developer and maintainer I wish to have. Feel free to promote them to the PR message if agree.

* As developer:

- [ ] I could expect to have offline documentation. This is any published format should not require resource loading from Internet.
- [ ] I don't expect the documentation to be shipped with production code. In general I always go to developer site for search the documentation. So docs folder and related file should be `exclude-ignore` in `.gitattributes`
- [X] I expect to contribute to documentation without to learn how to contribute first. Markdown is a good choice here
- [ ] Documentation is not only static characters. The option of lunch an example in a playground should be available. (May with extensions)
- [ ] Seems the render tool should display line numbers (It has been a recurring change in the actual documentation)
- [ ] Write documentation based in recipes and introduction by component. Personally I found to write formal documentation listing class method signatures is like to repeat the API docs.
- [ ] As reminder GitHub use a custom Markdown syntax named [GitHub Flavored Markdown](https://help.github.com/articles/github-flavored-markdown/) 

* As maintainer:

- [ ] Tool for validate syntax
- [ ] Continuos deployment
- [ ] Easy to compare a rendered version before and after
- [ ] Given PR waiting for documentation. Will be rejected if contributor don't attach the docs? 
- [ ] Document PHP files too. Many times I don't understand the purpose of many classes in the framework. Docblocks should be required too, in special class level docblock.
- [ ] Before to discuss how to deal with documentation in multiple locations, I would like to see how the framework (zf2 repo) will glue the new by component repo structure
- [ ] Release a compiled documentation from multiple location documentation is not compatible with independent livecycles (at least the compiled version won't be in sync with component versions)

weierophinney:
Rtfd.org used MkDocs, which has a number of issues:

- order is based on filename, which makes inserting new pages or reordering
painful.
- no way to import docs from other sources.
- templates are relatively difficult to customize (IMO).
- compilation requires an additional language + libraries to execute (no
essay way to render locally for casual contributors).

Maks3w:
> - no way to import docs from other sources.

RTFD has a bit support for subprojects

doc/book/api.md
```diff
@@ -0,0 +1,194 @@
+API
+===
+
+## Request Message
+
+`Zend\Diactoros\Request` implements `Psr\Http\Message\RequestInterface`, and is
```
Maks3w: I prefer if line length match visible GH space. 80 characters limit is too few (I hope any text editor will reflow before to send to print).

doc/book/api.md
```diff
@@ -0,0 +1,194 @@
...
+```php
+// Returns new ServerRequest instance, using values from superglobals:
+$request = ServerRequestFactory::fromGlobals();
+
+// or
+
+// Returns new ServerRequest instance, using values provided (in this
+// case, equivalent to the previous!)
+$request = RequestFactory::fromGlobals(
+  $_SERVER,
```
Maks3w: 4 spaces indentation.

doc/book/api.md
```diff
@@ -0,0 +1,194 @@
...
+  $_GET,
+  $_POST,
+  $_COOKIE,
+  $_FILES
+);
+```
+
+## URI
+
+`Zend\Diactoros\Uri` is an implementation of `Psr\Http\Message\UriInterface`,
```
Maks3w: hyperlink to UriInterface

README.md
```diff
@@ -14,362 +14,15 @@ This package supercedes and replaces [phly/http](https://github.com/phly/http).
...
-At times, it's useful to either create a string representation of a message (serialization), or to cast a string or stream message to an object (deserialization). This package provides features for this in `Zend\Diactoros\Request\Serializer` and `Zend\Diactoros\Response\Serializer`; each provides the following static methods:
-
-- `fromString($message)` will create either a `Request` or `Response` instance (based on the serializer used) from the string message.
-- `fromStream(Psr\Http\Message\StreamInterface $stream)` will create either a `Request` or `Response` instance (based on the serializer used) from the provided stream.
-- `toString(Psr\Http\Message\RequestInterface|Psr\Http\Message\ResponseInterface $message)` will create either a string from the provided message.
-
-The deserialization methods (`from*()`) will raise exceptions if errors occur while parsing the message. The serialization methods (`toString()`) will raise exceptions if required data for serialization is not present in the message instance.
-
-
-  [Master]: https://travis-ci.org/zendframework/zend-diactoros
```
Maks3w: This lines are required with the @gianarb way of inject the badges.

weierophinney:
That doesn't help us when we need to render them for the website, however.

doc/book/api.md
weierophinney: This was literally cut-and-paste from the old readme. Clearly, pushing docs
into their own directory and files prompts more thorough review! :-)

weierophinney:
> As developer:
> 
>  - [ ] I could expect to have offline documentation. This is any published format should not require resource loading from Internet.

Writing in markdown provides this, and writing with a tool that can be executed without additional tooling does as well. This is why couscous and bookdown were of interest to me; they install via composer, and can compile docs to HTML for offline viewing. If you don't want to run the tool, you can read and understand the markdown sources.

> - [ ] I don't expect the documentation to be shipped with production code. In general I always go to developer site for search the documentation. So docs folder and related file should be export-ignore in .gitattributes

I'll update the .gitattributes to reflect this; simply forgot.

> - [X] I expect to contribute to documentation without to learn a new language first. Markdown is a good choice here
> - [ ] I expect to contribute to documentation without to learn how to contribute first. "Fork me in GitHub" badge or "Edit in GitHub" link generation

When shipping docs with the code, we have this out-of-the-box. The situation with markdown is even better than we had with rst, due to the fact that they can preview a rendered version as they type, and see that it's semantically/structurally correct

For the manual, it becomes a tad bit more complex, but can likely be done with some creative templating and/or naming conventions in the documentation. Bookdown allows us to assign names to source files, which will be how the HTML files are named; if we use the component name as a prefix, we can tie into that with the templates.

> - [ ] Documentation is not only static characters. The option of launch an example in a playground should be available. (May with extensions)

This would be trivial; CommonMark (the PHP library that bookdown uses to parse and render markdown to HTML) allows templating; we could easily alter the generated output to have a link/JS that will submit to a service like 3v4l.org.

> - [ ] Seems the render tool should display line numbers (It has been a recurring change in the actual documentation)

I used highlight.js, a JS library, for the syntax highlighting; it has a runtime option you can use to specify this. Again, I simply forgot.

> - [ ] Write documentation based in recipes and introduction by component. Personally I found to write formal documentation listing class method signatures is like to repeat the API docs.

Again, the documentation in this PR is cut-and-paste from the README.md file, and is nothing new; I never intended to suggest documentation structure with this PR, only a potential toolchain.

In this particular case, I mainly demonstrated where the classes had API additions, such as the constructors. This is pertinent for somebody reading the documentation, as they can see then how to construct objects. That said, I also agree that API documentation likely should not be duplicated in formal documentation.

On the flip side, as a consumer, I also resent having to click through to a completely different tool with completely different layout and syntax highlighting and navigation when I want to look up signatures. IIRC, bookdown has a tool for auto-generating markdown for API documentation from your code, which could solve that particular gripe of mine. We would need to enforce updating such docs with merges, however.

> - [ ] As reminder GitHub use a custom Markdown syntax named GitHub Flavored Markdown

It does, but CommonMark (the parser/renderer bookdown uses) essentially supports it out-of-the-box already. This means contributors can use the syntax with which they are already familiar.

> As maintainer:
> 
> - [ ] Tool for validate syntax

Do you mean for the documentation itself, or for the code samples?

For the documentation, we've never had such a tool, and if you edited on GitHub itself via their code editor interface, it actually hindered your ability to understand the changes you made. For those using IDEs or editors, your ability to syntax check was based on your tools and their flexibility. In Vim, I use syntastic, which has support for RST, but which does not understand most of the extensions we use, even ones as trivial as `code-block`, which meant I'd see a ton of errors on *any* page I edited. The support for Markdown, however, because it has a smaller syntax, is solid, and has been in every other environment I've tried as well.

If you mean for code samples, this requires tooling, plain and simple. We've never had this, in DocBook, or RST.

>  - [ ] Continuos deployment

This can be relatively easy to implement no matter the solution, but far easier if the tooling is supported by the CI environment already. I am unsure if the PHP environment on Travis has support for python as well, but *do* know we can use PHP and composer, which is a point to consider (and a point in favor of couscous, daux.io, and bookmark.io).

Part of this, however, requires us making changes to the ZF website to simplify and accommodate such deployment; right now, it's a fairly painful, manual proces.

> - [ ] Easy to compare a rendered version before and after

This is another reason to use Markdown, as you *can* preview side by side in GitHub currently.

> - [ ] Given PR waiting for documentation. Will be rejected if contributor don't attach the docs?

That is the desired workflow, yes, and why I and others would like documentation for components to live *with* the components.

> - [ ] Document PHP files too. Many times I don't understand the purpose of many classes in the framework. Docblocks should be required too, in special class level docblock.

We're getting better about doing this, and I know in particular that @Ocramius is a stickler for this. If we make this part of the maintainer guidelines, a point in the CONTRIBUTING.md, and a rule for the code sniffer, we can enforce it consistently.

> - [ ] Before to discuss how to deal with documentation in multiple locations, I would like to see how the framework (zf2 repo) will glue the new by component repo structure

These are separate discussions. There are pertinent and relevant reasons for the approach I'm suggesting. For one, with tools like pandoc, we can actually take the compiled HTML from the full manual (aggregating all components) and create PDF and EPUB manuals. As such, how the ZF2 repo structure works, and how the ZF site works, are parallel topics, but not dependencies.

> - [ ] Release a compiled documentation from multiple location documentation is not compatible with independent livecycles (at least the compiled version won't be in sync with component versions)

It *is* relevant for a *framework*, release, however, which will be pinned to specific versions of each component. A manual for the *framework* would likely include the documentation for each component it depends on, as they form the full, public API. It will likely have additional pieces (getting started guide, tutorials, etc.), too.

I hope that addresses the bulk of your feedback. As you'll note, essentially every bullet point you raise is already addressed or is easily addressed using the proposed solution.

README.md
weierophinney: I'll reinstate those.

weierophinney:
> In my opinion we can use only read the docs :) no multiple location :)

We need to be able to have the docs on the website, in the same layout as the website; bouncing to an alternate site is a disconnect for users, and creates confusion. Additionally, we need to be able to export downloadable documentation, in HTML, PDF, and ePub formats.

weierophinney:
> Seems the render tool should display line numbers (It has been a recurring change in the actual documentation)
> I used highlight.js, a JS library, for the syntax highlighting; it has a runtime option you can use to specify this. Again, I simply forgot.

Forgot, but highlight.js does *not* offer line number support. There are other alternatives, however, that work with the exact same markup, and I can investigate those.

weierophinney:
@Maks3w I've incorporated most of your feedback at this time; will look into alternate syntax highlighters tomorrow.

harikt:
@weierophinney may be you need http://prismjs.com/plugins/line-numbers/

doc/templates/body.php
```diff
@@ -0,0 +1,7 @@
+<body>
+<div class="container-fluid">
+    <?php echo $this->render('core'); ?>
+</div>
+<script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/8.6/highlight.min.js"></script>
```
Maks3w: Use schema less URIs require to visualize documentation using a webserver, even for visualize the generated static HTML.

Preferable to use `https` always

doc/book/api.md
```diff
@@ -0,0 +1,186 @@
+API
+===
```
Maks3w: I prefer to use `# API` instead for to have a common syntax in the whole document

Maks3w:
`RTFD` or `GitHub Pages` does not require the user to bounce another site.

Contents can be served using CNAMEs or with a proxy module in the Zend web server

akrabat:
The only thing I care about is that we go back to the ZF1 way where new work contains code, tests and docs in a single patch.

gianarb:
@akrabat recaps my thought! ReadTheDocs is very easy to use and it is as a service.. but if exists another php project can be extended easy is a good solution :)

manuakasam:
@Maks3w first off: your list is very well written. I strongly sympathise
with the points you have highlighted. Especially this part:

>  Write documentation based in recipes and introduction by component.
Personally I found to write formal documentation listing class method
signatures is like to repeat the API docs.

I might be slightly biased on this, but a very good example for what you
want would likely be ZfcRbac
https://github.com/ZF-Commons/zfc-rbac/tree/master/docs


And i do agree with @akrabat, too. Which - as far as I understand - is the
now targeted approach again. I'm really excited for this. Documentation is
always the one thing that really puts people off of ZF (even though it
isn't as bad as it resides in most peoples minds...).

Rob Allen <notifications@github.com> schrieb am Mi., 27. Mai 2015 um
08:36 Uhr:

GeeH:
I actually disagree with one fundamental point that @Maks3w makes: 

> I don't expect the documentation to be shipped with production code. In general I always go to developer site for search the documentation. So docs folder and related file should be export-ignore in .gitattributes

I really would like the docs of each component shipped with the code with the caveat that it should be easy to checkout  a component without the docs.

If I am using the components as individual components, having the docs bundled with them is easy enough, and if I can checkout without the docs on production then so much the better. The whole purpose of splitting the repositories is so that the components can be used easier, and in my experience having the docs bundled with the code makes life SO much easier in this environment - I believe this is what most successful libraries do (granted, I only checked a few - see [Guzzle](https://github.com/guzzle/guzzle)). I'm guessing this is what @weierophinney alludes to with the `.gitignore` comment.

Maks3w:
@GeeH The reason for not ship docs with code are two.

1) Indirect dependencies (2nd grade of dependencies or greater) will download the documentation but is not expected to be used by the application/library under development

2) It's basically the same reason why we exclude so many things in production tarballs. Docs are only used at development time.

3) We expect to have friendly readable documentation. This is, the HTML version. At least between HTML and RST I preffer HTML version whenever possible.

Maks3w:
Bonus:

I have excluded vendor folder from my IDE (too noisy). I hate when I have to open the folder for to debug some unexpected behavior from my depedencies.

> If you need to look inside of your `vendor` folder then means your library provider does not provide enough error messages or there are missing Docblock documentation

gianarb:
@Maks3w html docs into the vcs? Maybe we might think a tool that build runtime docs and start server.. for example godoc in golang :)

GeeH:
What we are talking about here is personal opinion. You’d prefer not to look in the vendor folder whereas I don’t mind. It’s an opinionated decision which is the reason I think it’s not a valid point for discussion. You’re implying that these things are “fact” (at least to me, it might be a mis-communication) whereas I think that these things are just your point of view.

manuakasam:
The actual question that should be answered is but one:

Is there ANY negative impact of shipping the documentation with the sources?

To me, there's only positive gains in doing so (i.e. having the docs
directly available within github when i check out something).
But there may be negative aspects that I'm simply not seeing.

For the sake of discussion I'd argue that we should find out what actual
negative effects this could raise and if those are fixable ones or not.
Because if there's no negative aspects to this then the only answer should
be yes. Shouldn't it?

Gary Hockin <notifications@github.com> schrieb am Mi., 27. Mai 2015 um
12:38 Uhr:

Maks3w:
If HTML should be included in the repo then should be commited against github pages branch. Which is the same as don't ship in the tarball

weierophinney:
@harikt —

> @weierophinney may be you need http://prismjs.com/plugins/line-numbers/

Yes, likely will switch to prism (currently using highlight.js).

froschdesign:
> We need to be able to have the docs on the website, in the same layout as the website; bouncing to an alternate site is a disconnect for users, and creates confusion.

:+1:

(I think, with Bookdown it would be very easy to implement the documentation in the website. We can overwrite the default templates and add our HTML structure.)

weierophinney:
I've swapped the syntax highlighting from highlight.js to prism.js. I had to create a custom version in order to enable line numbers, which I'm currently hosting on S3; we can move that to the ZF site later if desired.

New screenshots:

api - google chrome_011

and mobile:

api - google chrome_012

weierophinney:
> The actual question that should be answered is but one:
> 
> Is there ANY negative impact of shipping the documentation with the sources?

What I'm seeing debated is not whether the documentation should be inside the component repositories, but rather:

- Should it be shipped inside production packages (i.e., the distribution packages composer uses for stable tags and/or when `--prefer-dist` and/or `--no-dev` are selected)?
- Should the _rendered_ version of the documentation be included in _any_ packages?

My take is as follows:

- Rendered documentation should only be included *if* the tool/process we choose cannot be run via PHP. We cannot expect people to have tools like pandoc or mkdoc available, or even alternate language runtimes. PHP, however, can be considered a reasonable baseline. For this reason, tools like couscous, daux.io, and bookdown are preferred, as they enable any consumer to render the documentation. If we go with another tool, it will be up to the maintainer to run the tool and commit the rendered documentation when merging.
- I have no preference as to whether documentation is in the production packages or not. I think there are reasonable arguments on either side. The reason we moved the ZF documentation to a separate repository was because it nearly tripled the size of the repo due to having 6 different languages and covering the entire framework. For individual components, my guess is that they will inflate the size, but not dramatically; we're talking a few kilobytes.

stefanotorresi:
Could we please continue the debate in zendframework/zf2-documentation#1489 ?
I completely missed this thread because I didn't expect this discussion to happen in this repo! :-P

weierophinney:
I've rebased, and updated the patch to:

- Remove templates (now in [zendframework/documentation](https://github.com/zendframework/documentation)).
- Remove CSS (now in [zendframework/documentation](https://github.com/zendframework/documentation)).
- Move `bookdown.json` up a directory (makes more sense to have it a level up, as it "governs" that directory).
- Made all headings use the `#` notation.

A note to those who were suggesting rtfd.org: this is *still* a possibility, and not mutually exclusive to using Bookdown. We can easily write a tool that could create the YAML configuration necessary for RTFD to render the markdown documents from the `bookdown.json` file. I'm just not convinced the rendered documentation will be necessary once we have the docs in markdown, as github does a reasonable job of rendering them.

README.md
weierophinney: Fixed in the most current version.

gianarb:
Two different documentation deliveries are very difficult to understand and maintain.. :) IMO..
If bookdown is good we can use it and stop :)

CONTRIBUTING.md
```diff
@@ -34,6 +34,15 @@ to upgrade and/or update in order to protect their applications.
 
 For sensitive email communications, please use [our PGP key](http://framework.zend.com/zf-security-pgp-key.asc).
 
+## Documentation
+
+Documentation is in [GitHub Flavored Markdown](https://help.github.com/articles/github-flavored-markdown/),
```
colinodell: Doesn't bookdown use CommonMark (instead of GFM)?

weierophinney: It does, but it renders all GFM syntax, and that's the point we want to get
across (particularly as the default rendering location is on github).

colinodell: > It does, but it renders all GFM syntax

Well, there are a couple GFM features not supported by CommonMark:

 - Strikethrough (`~~example~~`)
 - Tables

I get where you're coming from though.

PR #31, opened 2015-05-31
Fix: Diactoros has actually just one dependency

localheinz:
This PR

* [x] fixes `README.md`, as `zendframework/zend-diactoros` actually has just one dependency, not many


Maks3w:
I think the paragraph need to be rewritten. Does not have sense to have a list with only one element.

localheinz:
@Maks3w 

:+1:

PR #36, opened 2015-05-31
Fix: Add property annotations to class-level docblock

localheinz:
This PR

* [x] adds `@property` annotations for fields which seem to be declared dynamically in `RequestTrait`, but are apparently declared in `MessageTrait`

:bulb: Not sure if this is right - is it?

### DIFF
--- a/src/RequestTrait.php
+++ b/src/RequestTrait.php
@@ -23,6 +23,10 @@ use Psr\Http\Message\UriInterface;
  * the enviornment. As such, this trait exists to provide the common code
  * between both client-side and server-side requests, and each can then
  * use the headers functionality required by their implementations.
+ *
+ * @property array $headers
+ * @property array $headerNames
+ * @property StreamInterface $stream
  */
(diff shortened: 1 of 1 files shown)


Maks3w:
May this reveal a coding design issue.

localheinz:
@Maks3w 

To me it reveals that the two traits can't be used independently - or one of them is missing declared fields, no?

Maks3w:
exactly. This require an in depth analysis.

localheinz:
@Maks3w 

:+1:

weierophinney:
> To me it reveals that the two traits can't be used independently

They can't. However, PHP does not offer trait inheritance.

The alternate solution to this would be to use abstract class instead of traits, but that would present a BC break. I'm also not terribly fond of abstract classes.

I'll merge this as-is for now, and we can discuss the larger issue at a later date.

localheinz:
@weierophinney 

:+1:

PR #28, opened 2015-05-27
[] short array syntax

samsonasik:

PR #38, opened 2015-05-31
[cs] Place CS rules in config file

Maks3w:
Supersede and close https://github.com/zendframework/zend-diactoros/pull/28/

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
   },
   "require-dev": {
     "phpunit/PHPUnit": "~4.6",
-    "squizlabs/php_codesniffer": "~2.0"
+    "squizlabs/php_codesniffer": "^2.3.1"
   },
   "provide": {
     "psr/http-message-implementation": "~1.0.0"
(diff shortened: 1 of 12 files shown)


localheinz:
@Maks3w 

:+1:

composer.json
```diff
@@ -25,7 +25,7 @@
   },
   "require-dev": {
     "phpunit/PHPUnit": "~4.6",
-    "squizlabs/php_codesniffer": "~2.0"
+    "squizlabs/php_codesniffer": "^2.3.1"
```
weierophinney: Just curious, why this specific version and up? What changed specifically with 2.3.1 that is needed?

Maks3w: long array rule not exists before 2.3.1

Maks3w: Yea they didn't follow semantic versioning about to raise feature number when new sniffs are added

weierophinney: Okay, cool; was just curious. :)

PR #47, opened 2015-06-04
improve header validation speed

mtymek:
Replacing `for` loop with regular expressions brought visible performance improvement to header validation.

See simple benchmark: https://gist.github.com/mtymek/660b388db0c5b9d63bc0

### DIFF
--- a/src/HeaderSecurity.php
+++ b/src/HeaderSecurity.php
@@ -106,23 +106,15 @@ final class HeaderSecurity
             return false;
         }
 
-        $length = strlen($value);
-        for ($i = 0; $i < $length; $i += 1) {
-            $ascii = ord($value[$i]);
-
-            // Non-visible, non-whitespace characters
-            // 9 === horizontal tab
-            // 10 === line feed
-            // 13 === carriage return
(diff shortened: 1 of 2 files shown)


src/HeaderSecurity.php
```diff
@@ -106,23 +106,15 @@ public static function isValid($value)
...
-                return false;
-            }
+        // Non-visible, non-whitespace characters
+        // 9 === horizontal tab
+        // 10 === line feed
+        // 13 === carriage return
+        // 32-126, 128-254 === visible
+        // 127 === DEL (disallowed)
+        // 255 === null byte (disallowed)
+        if (preg_match('/[^\x09\x0a\x0d\x20-\x7E\x80-\xFE]/', $value)) {
```
Maks3w: extract the regex as `static $regex = '....'` improve something?

mtymek: I don't see any difference.

weierophinney: There won't be a difference. PHP's PCRE caches regexen that it encounters, so subsequent matches using the same expressions will not need re-compilation.

PR #48, opened 2015-06-04
Drop minimum required PHP version to 5.4

weierophinney:
Drupal 8 has a minimum PHP of 5.4.8. Symfony 2.7 still depends on 5.3.something. 5.4 is a reasonable middle-ground to ensure that new libraries targeting older versions of PHP do not need to be written.

This will be a temporary change; we will likely update the minimum required version in one of the next few minor releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ cache:
 matrix:
   fast_finish: true
   include:
+    - php: 5.4
     - php: 5.5
       env:
         - EXECUTE_CS_CHECK=true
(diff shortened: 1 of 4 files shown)


Crell:
This would eliminate one of the two blockers to Drupal pulling in the Symfony-PSR-7 bridge and Diactoros, so definitely +1 from me.  (The other blocker is a question of scope and timing, which is not something Zend can do anything about.)

dunglas:
It will increase the number of Symfony users and install able to use the PSR-7 Bridge so :+1: 
Thanks!

PR #51, opened 2015-06-08
Fix MessageInterface::getHeaderLine contract violation

Maks3w:


test/MessageTraitTest.php
```diff
@@ -200,7 +200,7 @@ public function testGetHeaderReturnsAnEmptyArrayWhenHeaderDoesNotExist()
 
     public function testGetHeaderLineReturnsNullWhenHeaderDoesNotExist()
```
weierophinney: Change the name, too, `s/Null/EmptyString/`.

test/RequestTest.php
```diff
@@ -344,7 +344,7 @@ public function testGetHostHeaderLineReturnsUriHostWhenPresent()
     public function testGetHostHeaderLineReturnsNullIfNoUriPresent()
```
weierophinney: Change the name, too, `s/Null/EmptyString/`

Maks3w: done

PR #55, opened 2015-06-14
ResponseTrait

mrgrain:
Moved related methods to a ResponseTrait
Using ResponseTrait, a class will implement the ResponseInterface
This basically allows for creating custom responses without the need to use the default Response constructor.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -22,87 +22,7 @@ use Psr\Http\Message\StreamInterface;
  */
 class Response implements ResponseInterface
 {
-    use MessageTrait;
-
-    /**
-     * Map of standard HTTP status code/reason phrases
-     *
-     * @var array
-     */
-    private $phrases = [
-        // INFORMATIONAL CODES
-        100 => 'Continue',
-        101 => 'Switching Protocols',
-        102 => 'Processing',
-        // SUCCESS CODES
-        200 => 'OK',
-        201 => 'Created',
-        202 => 'Accepted',
-        203 => 'Non-Authoritative Information',
-        204 => 'No Content',
-        205 => 'Reset Content',
-        206 => 'Partial Content',
-        207 => 'Multi-status',
-        208 => 'Already Reported',
-        // REDIRECTION CODES
-        300 => 'Multiple Choices',
-        301 => 'Moved Permanently',
-        302 => 'Found',
-        303 => 'See Other',
-        304 => 'Not Modified',
-        305 => 'Use Proxy',
-        306 => 'Switch Proxy', // Deprecated
-        307 => 'Temporary Redirect',
-        // CLIENT ERROR
-        400 => 'Bad Request',
-        401 => 'Unauthorized',
-        402 => 'Payment Required',
-        403 => 'Forbidden',
-        404 => 'Not Found',
-        405 => 'Method Not Allowed',
-        406 => 'Not Acceptable',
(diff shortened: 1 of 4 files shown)


weierophinney:
The main reason `RequestTrait` exists is because there are two `RequestInterface` implementations (`Request` and `ServerRequest`), and I didn't want them to inherit from each other so that I could better protect encapsulation.

Can you demonstrate some of the use cases you have for this change, please?

mrgrain:
Yes sure, for a example a `RedirectResponse` could be created as simple as:
```php
class RedirectResponse
{
    use ResponseTrait;

    public function __construct($redirectUrl)
    {
        $this->stream     =  new Stream('php://memory', 'wb+');
        $this->statusCode = 302;
        $this->headers = ['Location: '.$redirectUrl];
    }
}
```
This is still not perfect, as a lot has to be done manually (e.g. setting the reason phrase) but it offers an other approach for custom responses, in addition to the static generator methods from the develop-branch.

Maks3w:
@mrgrain I think inheritance it's better in your example. Diactotors Response and your example RedirectResponse share the same semantic field

mrgrain:
@Maks3w Yes and no. I would always prefer inheritance. The issue I have with Diactotors Response is it's signature. Especially the stream string/resource as first parameter is something I really don't care about. I know this is an important and interesting value to have, but in most cases I actually would consider the stream an implementation detail.

weierophinney:
@mrgrain One interesting fact of PHP is that constructors can differ in child classes. This allows you to do the following:

```php
class Response implements ResponseInterface
{
    public function __construct($body = 'php://memory', $status = 200, array $headers = [])
    {
        /* initialize */
    }
}

class Redirect extends Response
{
    public function __construct($redirectUrl, $status = 302)
    {
        parent::__construct(
            'php://memory',
            $status,
            [
                'location' => [$redirectUrl],
            ]
        );
    }
}

```

PHP will raise no errors in the above; constructors are exempt from the parameter inheritance rules present for any other method.

The problem that *does* arise, however, is that we've defined all properties in our messages as private. This is to enforce encapsulation. It does mean, however, that if you want to extend the messages, you cannot directly manipulate properties without first redefining them in your extending class — and this may lead to odd issues later down the line. And this, @Maks3w, is why inheritance may not be the correct solution for all use cases.

That said: the example I provided above already works, and doesn't require having a trait. Knowing this, @mrgrain, is the PR necessary?

weierophinney:
@mrgrain BTW, you might be interested in the following:

- [Documentation for #58 and #52](https://github.com/zendframework/zend-diactoros/blob/develop/doc/book/custom-responses.md), which covers custom responses (I wrote it with this PR in mind).
- #59, which creates a `RedirectResponse`, similar to how I demonstrated above.

weierophinney:
I'm going to close this, as I think I've sufficiently demonstrated that you can create custom extensions trivially, and I'm not convinced of the need for extracting the response-specific functionality into a trait at this time.

mrgrain:
Yes, that's perfectly fine for me :) I amazingly didn't know about the possibility to overload constructors. Learned something new today.

Thanks for the documentation links. They are looking very promising.

weierophinney:
@mrgrain Those will be in the upcoming 1.1 version; I'll likely tag it in the next couple of days. :)

PR #57, opened 2015-06-19
File validation fails when file contains sub-arrays

jonhudson:


weierophinney:
Please update the pull request summary to describe what this accomplishes; since I link to the PRs in the changelog, that's often the first place people look to understand how the change impacts them.

To be clear: we already support sub-arrays, but specifically _indexed_ sub-arrays, vs _associative_ sub-arrays, which is what this PR addresses.

src/ServerRequestFactory.php
```diff
@@ -469,6 +469,7 @@ private static function createUploadedFileFromSpec(array $value)
      */
     private static function normalizeNestedFileSpec(array $files = [])
     {
+        $normalizedFiles = [];
```
weierophinney: There's no need to create a new array here.

weierophinney: I stand corrected when running the unit tests. Still trying to understand why this change results in any change in behavior…

weierophinney:
I've updated the summary for you.

src/ServerRequestFactory.php
jonhudson: The `UploadedFile` objects were being appended to the `$files` array, which meant that the `tmp_name` and other attributes preceeded them in the array. When same array gets passed to the validation function, it was trying to then validate strings such as the `tmp_name` as an instance of `UploadedFile`, and failing. So I just created a new array to store the `UploadedFile` objects in and returned that instead.

jonhudson:
Thank you.

PR #58, opened 2015-06-20
Add shortcut method for empty responses

franzliedke:
`StringResponse::blank(401)`, as that's another common use case (though now `StringResponse` may not be perfect).

It's called `blank()` because `empty()` cannot be used as method name until PHP 7. Any better ideas?

### DIFF
--- a/src/Response/StringResponse.php
+++ b/src/Response/StringResponse.php
@@ -61,6 +61,18 @@ final class StringResponse
     }
 
     /**
+     * Create an empty response with the given status code.
+     *
+     * @param int $status Status code for the response, if any.
+     * @param array $headers Headers for the response, if any.
+     * @return Response
+     */
+    public static function empty($status = 200, array $headers = [])
+    {
+        return static::createResponse('', $status, $headers);
+    }
+
+    /**
      * This class is non-instantiable.
      */
     private function __construct()
@@ -71,7 +83,7 @@ final class StringResponse
      * Create a Response from the provided information.
      *
      * Creates a Response using a php://temp stream, and writes the provided
-     * body to the stream; if non content-type header was provided, the given
+     * body to the stream; if no content-type header was provided, any given
      * $contentType is injected for it.
      *
      * @param string $body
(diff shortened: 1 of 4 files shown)


weierophinney:
@franzliedke I'd argue a status 204 with a reason phrase "No Content" as defaults would make most sense here; you could then call the method `noContent()`. Thoughts?

src/Response/StringResponse.php
```diff
@@ -80,12 +92,12 @@ private function __construct()
      * @param string $contentType
      * @return Response
      */
-    private static function createResponse($body, $status, array $headers, $contentType)
+    private static function createResponse($body, $status, array $headers, $contentType = null)
```
weierophinney: Keep the original signature, but have the caller send a null value or an empty string.

weierophinney:
Also, needs tests. :)

franzliedke:
@weierophinney Took care of your feedback:
Parameter restored, method renamed, status code 204 by default, test added.

weierophinney:
On review, I'm not sure this belongs in the `StringResponse`; there's no string!

I was commenting earlier on another PR, and I wonder if perhaps we could do this instead:

```php
namespace Zend\Diactoros\Response;

use Zend\Diactoros\Response;
use Zend\Diactoros\Stream;

class NoContent extends Response
{
    public function __construct(array $headers = [], $status = 204)
    {
        $body = new Stream('php://memory', 'r');
        parent::__construct($body, $status, $headers);
    }
}
```

With usage like:

```php
// Assuming "use Zend\Diactoros\Response\NoContent;
return new NoContent();
```

This provides the simplicity you're looking for, while keeping it semantically separate from the string-based response types. The constructor omits the body, and ensures the body is read-only. Since a "no content" status typically won't change, headers become the first argument, as those are the one aspect of the response that you may want to vary (e.g., to provide a location header, a link header, an authorization token, etc.).

Thoughts?

franzliedke:
Why not.

I wouldn't change the arguments order, though, that would be confusing. Also, you may want to have empty responses for something like `201 Created` or `202 Accepted`, too, so changing the status is not so uncommon.

When only wanting to change the header, but not caring about the status, we could even provide an alternate static constructor, or people could just resort to this:

~~~php
$response = (new NoContent)->withHeaders(...);
~~~

weierophinney:
@franzliedke You've convinced me; go for it!

franzliedke:
Can we call it `EmptyResponse` now?

weierophinney:
@franzliedke sure.

franzliedke:
Done as you suggested. Used `php::temp` instead, though, as that's what we had in the `StringResponse` class.

weierophinney:
@franzliedke I'd chosen `php://memory` as the liklihood of an empty response ever going to the temp filesystem is essentially nil. :) Either will work, though; `php://temp` uses memory until memory limits are reached.

franzliedke:
Ah, thanks for clarifying.

I think I'm done with this one. :)

PR #59, opened 2015-06-23
Implement RedirectResponse

weierophinney:
Besides the HTML and JSON responses introduced in #52 and the empty response implemented in #58, the other most common server-side response type is a redirect. This pull request implements a `RedirectResponse`, which accepts a URI in order to produce a 302 response with a `Location` header. As per the other response types, this custom response type allow specifying an alternate status code and headers to use at initialization.

Examples as documented:

```php
// 302 redirect:
$response = new RedirectResponse('/user/login');

// 301 redirect:
$response = new RedirectResponse('/user/login', 301);

// using a URI instance (e.g., by altering the request URI instance)
$uri = $request->getUri();
$response = new RedirectResponse($uri->withPath('/login'));
```


weierophinney:
ping @franzliedke — thought you might find this one interesting.

franzliedke:
Cool. I like it. :+1:

PR #29, opened 2015-06-24
Custom HTTP method

weierophinney:
It will be nice feature to define own custom http method. Now it's not possible to override or add new one. I can provide PR, but I want to know what are you think about it.

### DIFF
--- a/src/RequestTrait.php
+++ b/src/RequestTrait.php
@@ -48,23 +48,6 @@ trait RequestTrait
     private $uri;
 
     /**
-     * Supported HTTP methods
-     *
-     * @var array
-     */
-    private $validMethods = [
-        'CONNECT',
-        'DELETE',
-        'GET',
-        'HEAD',
-        'OPTIONS',
-        'PATCH',
-        'POST',
-        'PUT',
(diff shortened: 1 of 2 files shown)


src/RequestTrait.php
```diff
@@ -48,23 +48,6 @@
...
-        'GET',
-        'HEAD',
-        'OPTIONS',
-        'PATCH',
-        'POST',
-        'PUT',
-        'TRACE',
-    ];
-
-    /**
```
weierophinney: Removed the above, as it's no longer necessary; validation is not against this list.

test/RequestTest.php
```diff
@@ -120,7 +120,7 @@ public function invalidRequestMethod()
             'false'      => [ false ],
             'int'        => [ 1 ],
             'float'      => [ 1.1 ],
-            'bad-string' => [ 'BOGUS-METHOD' ],
+            'bad-string' => [ 'BOGUS METHOD' ],
```
weierophinney: This test case had to change as the string was a valid token! Adding whitespace invalidated it again.

PR #64, opened 2015-06-25
Remove casting prior to JSON serialization

weierophinney:
As noted on #63, casting nulls and scalars to objects and arrays, respectively, is not necessary, and likely unwanted; it could even be considered buggy behavior.

This patch eliminates the casting entirely, and updates the tests to ensure that scalars are encoded verbatim.


franzliedke:
:+1:

weierophinney:
Somehow a 100% coverage dropped to 100% leading to a -0.03% change. Stupid coveralls!

Issue #65, opened 2015-06-29
UriInterface::__toString adds extra delimiters


nyamsprod:
I'm finishing coding League\Url v4 and it exposed the following class in public API
```php
Url::sameValueAs(Psr\Http\Message\UriInterface $url); 
```
To compare two PSR-7 UriInterface object based on their `__toString` method output.

I tried to make it work with Diactoros but I failed with the following edge cases URLs:
-  `http:/example.com`
-  `http:example.com`

In both cases the object is instantiated and the respective Uri components are correctly set but the `__toString` method adds extra path and/or authority delimiters.

Below the code to reproduce the bugs:

```php
use Zend\Diactoros\Uri;
$url = new Uri('http:/example.com');
echo $url->__toString(); //expected http:/example.com but got http:///example.com
new Uri($url->__toString()); //throw InvalidArgumentException
```
The authority delimiters are added (not required according to PSR-7) and the method returns an invalid Uri.

```php
use Zend\Diactoros\Uri;
$url = new Uri('http:example.com');
echo $url->__toString(); //expected http:example.com but got http:///example.com
new Uri($url->__toString()); //throw InvalidArgumentException
```
The authority delimiters and a path delimiter are added (not required according to PSR-7) and the method returns an invalid Uri.

weierophinney:
@nyamsprod Two questions:

- Is this worth correcting? As you yourself note, these are edge cases.
- If you feel it is worth correcting, how would you approach it?

A couple notes from the `UriInterface`:

- "If an authority is present, it MUST be prefixed by '//'". However, PSR-7 *does not* detail what to do if no authority is present. In other words, while the `//` is not *required* if no authority is present, it's up to the implementation to decide whether or not to add the delimiters.
- "If the path is rootless and an authority is present, the path MUST be prefixed by '/'" **and** "If the path is starting with more than one '/' and no authority is present, the starting slashes MUST be reduced to one." 

My takeaway from this (and, recall, I was *editor* of the spec) is that how Diactoros returns the URI is not unexpected; this is an edge-case. When I first looked at the URIs, before I examined what you presented them doing, I expected both would be spit back out as `http:///example.com`. This goes hand-in-hand with [our recent security disclosure](http://framework.zend.com/security/advisory/ZF2015-05), which would try and create a URI that will not have unintended XSS/redirect consequences.

With that information in mind, how would you proceed?

nyamsprod:
@weierophinney here's what I think

[...]

PR #66, opened 2015-07-01
Added additional tests to validate PSR-7 UriInterface implementation

nyamsprod:
The tests are added with an explanation in the docblock referencing PSR-7 and/or RFC3986 rules to explain the expected results.

These tests cover issue #65 I had submitted previously

IMHO none of the added tests should failed

### DIFF
--- a/test/UriTest.php
+++ b/test/UriTest.php
@@ -455,4 +455,101 @@ class UriTest extends TestCase
         $uri = new Uri($url);
         $this->assertEquals('http://example.org/zend.com', (string) $uri);
     }
+
+    /**
+     * As Per PSR7 UriInterface the host MUST be lowercased
+     *
+     * @group uriinterface
+     */
+    public function testHostnameMustBeLowerCasedAsPerPsr7Interface()
+    {
+        $url = 'http://WwW.ExAmPlE.CoM';
+        $uri = new Uri($url);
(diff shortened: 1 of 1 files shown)


weierophinney:
@nyamsprod Question: how do you parse the URI, if not using `parse_url()`? Are you using a dedicated parser in league/uri? What kind of performance overhead does that add? (We have a parser in zend-http, though, based on the number of bugs it has had logged to it over the years, I think the approach may be dubious at best.)

Part of the reason I ask is because addressing the test failures may require a substantial amount of work to address. If you know of a parser that could be used as a dependency, that may help us in getting these addressed sooner rather than later.

nyamsprod:
@weierophinney I'm using a dedicated parser for the package [UriParser](https://github.com/thephpleague/uri/blob/master/src/UriParser.php). It is based on the provided RFC3986 regular expression. I have yet to for its overhead though.

nyamsprod:
I've quickly made this [benchmark package](https://github.com/nyamsprod/uri-parser-benchmarks#overhead-capabilities) to test both URI parser implementations hope you'll find it interesting. The results on my computer shows that the speed depends on the URI structure. But memory usage seems to indicate that `League\Uri\UriParser` consume less than the `Zend\Uri\Uri` parser.

PR #71, opened 2015-07-28
Fixed docblock

bakura10:
Hi,

I think the docblock here is wrong, as you're doing a json_encode a bit later in the process.

### DIFF
--- a/src/Response/JsonResponse.php
+++ b/src/Response/JsonResponse.php
@@ -35,7 +35,7 @@ class JsonResponse extends Response
      * - JSON_HEX_AMP
      * - JSON_HEX_QUOT
      *
-     * @param string $data Data to convert to JSON.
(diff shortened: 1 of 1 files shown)


src/Response/JsonResponse.php
```diff
@@ -35,7 +35,7 @@ class JsonResponse extends Response
      * - JSON_HEX_AMP
      * - JSON_HEX_QUOT
      *
-     * @param string $data Data to convert to JSON.
+     * @param array $data Data to convert to JSON.
```
mwillbanks: Actually it could be anything but a resource.  so scalar|array|object :)

weierophinney: Considering the internal `jsonEncode()` method typehints as `mixed`, that's what I'll do on merge.

PR #73, opened 2015-07-29
Added vendor to travis cache

weierophinney:
- Per http://blog.wyrihaximus.net/2015/07/composer-cache-on-travis/

Mainly testing this out. :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ language: php
 cache:
   directories:
     - $HOME/.composer/cache
+    - vendor
 
 matrix:
   fast_finish: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Added
 
-- Nothing.
+- [#73](https://github.com/zendframework/zend-diactoros/pull/73) adds caching of
+  the vendor directory to the Travis-CI configuration, to speed up builds.
 
 ### Deprecated
 


weierophinney:
You can see the differences between the two builds:

- https://travis-ci.org/zendframework/zend-diactoros/builds/73249995
- https://travis-ci.org/zendframework/zend-diactoros/builds/73250365

HHVM does not benefit much from the change, but the other builds take around 50% less time overall, and the full time for all jobs dropped by a > 30%. Definitely worth it!

Maks3w:
1. We need a way of print what depenency versions are used. At this moment we are blind on that.
2. Take this other comparison
https://travis-ci.org/zendframework/zend-diactoros/builds/70634633
https://travis-ci.org/zendframework/zend-diactoros/builds/73250365

It's not fair to evaluate a cold cache 

I suggest to fix 1 and don't propagate this across the whole repos until we can verify how works when new versions are released

weierophinney:
@Maks3w (1) can be fixed as easily as adding this as a `before_script`:

```yaml
- composer info -i
```

Regarding (2), I honestly don't understand why the 1.1.2 release built so quickly; most of the other builds I've seen have been closer to the original build of this PR.

I'll get a PR in place with the `composer info -i` change.

Maks3w:
Travis runs faster in the morning (Europe) night (America) because the anonymous rate limit against GitHub API is not so exhausted.

If Composer can install using dist version (not clone) then the dependency can be sucesfully saved in the cache

IMO we have a very good decent speed rate. I hope Composer will improve the work with Travis-CI cache without the need of this kind of workarounds

PR #74, opened 2015-07-29
Print installed dependencies

weierophinney:
Per comment from @Maks3w on #73, this adds an `install` script to print the
installed dependencies.


weierophinney:
@Maks3w https://travis-ci.org/zendframework/zend-diactoros/jobs/73263512#L120

Maks3w:
:+1: Good. Without this we can't reproduce travis failures on our devices.

PR #77, opened 2015-08-04
Add a TextResponse to easily return plain text HTTP responses

### LINKED ISSUE #76: Add a TextResponse?
There is currently the base `Response`, but also `HtmlResponse`, `JsonResponse`, etc.

That's practical when we want to return HTML or JSON, but there is no simple `TextResponse` (or `PlainTextResponse`. That's not a huge deal, but writing a simple *hello world* is just plain confusing.

Example:

```php
// Super verbose
$response = new Response();
$res = $res->withHeader('content-type', 'text/plain');
$response->getBody()->write('Hello world!');
return $response;

// Wrong content type, but at least it's simpler and shorter
return new HtmlResponse('Hello world');
```

Do you think it would make sense to add a new `TextResponse` class?

Or even simpler if we could use the base Response class to return plain text content, but the $body parameter in the constructor seems to only accept streams. Maybe it could accept a plain string? (but then how to differentiate from a stream…)

mnapoli:
Fixes #76

I did not add a charset as the HtmlResponse didn't have one. Maybe this is something to add? (see e.g. [in symfony](https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/HttpFoundation/Response.php#L282-L288)). I am very far from an expert in HTTP stuff, so I'm just bringing it up but I have no idea (and it should probably be done in a separate PR anyway).

### DIFF
--- a/doc/book/custom-responses.md
+++ b/doc/book/custom-responses.md
@@ -23,6 +23,21 @@ Some standard use cases, however, make this un-wieldy:
 Starting with version 1.1, Diactoros offers several custom response types for simplifying these
 common tasks.
 
+## Text Responses
+
+`Zend\Diactoros\Response\TextResponse` creates a plain text response. It sets the
+`Content-Type` header to `text/plain` by default:
+
+```php
+$response = new TextResponse('Hello world!');
+```
+
+The constructor accepts two additional arguments: a status code and an array of headers.
+
+```php
+$response = new TextResponse($text, 200, ['Content-Type' => ['text/csv']]);
+```
+
 ## HTML Responses
 
 `Zend\Diactoros\Response\HtmlResponse` allows specifying HTML as a payload, and sets the
--- a/src/Response/TextResponse.php
+++ b/src/Response/TextResponse.php
@@ -1,0 +1,73 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @see       http://github.com/zendframework/zend-diactoros for the canonical source repository
+ * @copyright Copyright (c) 2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-diactoros/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Diactoros\Response;
+
+use InvalidArgumentException;
+use Psr\Http\Message\StreamInterface;
+use Zend\Diactoros\Response;
+use Zend\Diactoros\Stream;
+
+/**
+ * Plain text response.
+ *
+ * Allows creating a response by passing a string to the constructor;
+ * by default, sets a status code of 200 and sets the Content-Type header to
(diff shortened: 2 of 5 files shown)


doc/book/custom-responses.md
```diff
@@ -23,6 +23,21 @@ Some standard use cases, however, make this un-wieldy:
 Starting with version 1.1, Diactoros offers several custom response types for simplifying these
 common tasks.
 
+## Text Responses
+
+`Zend\Diactoros\Response\TextResponse` allows creating a plain text response. It sets the
+`Content-Type` header to `text/html` by default:
```
akrabat: text/plain, not text/html.

doc/book/custom-responses.md
```diff
@@ -23,6 +23,21 @@ Some standard use cases, however, make this un-wieldy:
 Starting with version 1.1, Diactoros offers several custom response types for simplifying these
 common tasks.
 
+## Text Responses
+
+`Zend\Diactoros\Response\TextResponse` allows creating a plain text response. It sets the
```
akrabat: replace "allows creating" with "creates".

src/Response/TextResponse.php
```diff
@@ -0,0 +1,73 @@
...
+ * Allows creating a response by passing a string to the constructor;
+ * by default, sets a status code of 200 and sets the Content-Type header to
+ * text/plain.
+ */
+class TextResponse extends Response
+{
+    use InjectContentTypeTrait;
+
+    /**
+     * Create an HTML response.
```
akrabat: Not HTML…

src/Response/TextResponse.php
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * Produces an HTML response with a Content-Type of text/plain and a default
```
akrabat: Not HTML.

mnapoli:
Thanks @akrabat I've updated the commit.

weierophinney:
> I did not add a charset as the HtmlResponse didn't have one. Maybe this is something to add?

Yes, please. :)

weierophinney:
Looks good; I'll wait for your feedback as to whether or not you want to add charset to this PR as well before merging, however.

mnapoli:
Will do tomorrow! I'll add it on HtmlResponse too. UTF-8 by default?

weierophinney:
> Will do tomorrow! I'll add it on HtmlResponse too.

Awesome!

> UTF-8 by default?

Of course!

mnapoli:
I've pushed a new commit (as I think this is a separate change since it also applies to HtmlResponse). 

From what I have read online (example: http://stackoverflow.com/a/19815111/245552) JSON is UTF-8 (mandatory) so there is no need for a charset.

weierophinney:
Merged to develop for release with 1.2.0. Thanks, @mnapoli !

PR #88, opened 2015-08-20
Set the Content-Length header when emitting the response

### LINKED ISSUE #84: Response emitter not setting Content-Length
This is more of a question about the response emitters than a bug report.

Slim does this processing before emitting the response: https://github.com/slimphp/Slim/blob/3.x/Slim/App.php#L472-L484

Should this be added to Diactoros, or is it expected to be done by the consuming project (i.e. me)?

mnapoli:
Fixes #84

One test is failing because of a side-effect in unrelated tests. Need to work on it, but if you have an idea on the best way to fix it's welcome, I'm not familiar with everything yet.

### DIFF
--- a/test/ServerTest.php
+++ b/test/ServerTest.php
@@ -13,6 +13,7 @@ use PHPUnit_Framework_TestCase as TestCase;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Zend\Diactoros\Server;
+use Zend\Diactoros\Stream;
 use ZendTest\Diactoros\TestAsset\HeaderStack;
 
 class ServerTest extends TestCase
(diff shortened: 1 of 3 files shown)


test/ServerTest.php
```diff
@@ -253,6 +254,16 @@ public function testListenPassesCallableArgumentToCallback()
...
+        $responseBody = new Stream('php://temp');
+        $this->response
+            ->expects($this->any())
+            ->method('getBody')
+            ->willReturn($responseBody);
+        $this->response
+            ->expects($this->any())
+            ->method('withHeader')
+            ->willReturnSelf();
+
```
mnapoli: Not really happy with that, mocking the whole world to get an integration test passing isn't really great. Didn't have a lot of time to think of a better solution.

mnapoli:
The tests should be green now.

Please review the tests and let me know if you want me to apply a different solution: auto-adding the `Content-Length` header broke several tests assertions (unexpected `Content-Length` header appeared).

test/ServerTest.php
weierophinney: With some of these integration tests, it seems like the only way to do it. They do get lengthy, though!

We might consider porting them to Prophecy at some point (we're using it extensively in the ZF3 component rewrites, and in Expressive), but that's something for a lull time in development.

Issue #70, opened 2015-07-23
Stream as callback


oscarotero:
Hi.
I have an application that does some heavy work and I'd like to send to the client a stream with the output of the process. I mean something like this:

```php
use Symfony\Component\Process\Process;

function handler ($request, $response) {
    $process = new Process($command);

    $stream = new StreamCallback(function () use ($process) {
        $process->run(function ($type, $buffer) {
            echo $buffer;
            flush();
        });
    });

    return $response->withBody($stream);
);
```

Is there any way to do something like this?

weierophinney:
We currently only have a single `StreamInterface` implementation, `Zend\Diactoros\Stream`. I'm interested in having additional implementations, and am open to pull requests to implement them.

I did some prototypes in response to questions while we were developing PSR-7, however, and they might interest you:

- https://gist.github.com/weierophinney/7c450a4ab3080c6a0413

oscarotero:
That prototypes looks great. I'll make a PR with a CallbackStream. Thank you

PR #90, opened 2015-08-25
added CallbackStream

oscarotero:
Hello.
This is a implementation of Stream that uses a callback as described in #70 
Maybe should be great to add support for generators, to allow doing thinks like this:
```php
$stream = new CallbackStream(function () {
    $command = new Command();
    $command->run();

    while ($command->isRunning()) {
        yield $command->getOutput(); //returns always a string
    }
    return false; //end of the command
});
```
In this case, the emitter should iterate with the stream, instead return all content at once. I mean:
```php
while (!$stream->eof()) {
    echo $stream->read(1);
};
```
This behaviour is not implemented (currently `$stream->read()` throws a `RuntimeException` and the callback can be executed only once), but I can do it if you like.

### DIFF
--- a/src/CallbackStream.php
+++ b/src/CallbackStream.php
@@ -1,0 +1,182 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @see       http://github.com/zendframework/zend-diactoros for the canonical source repository
+ * @copyright Copyright (c) 2015 Oscar Otero (http://oscarotero.com) / Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-diactoros/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Diactoros;
+
+use InvalidArgumentException;
+use RuntimeException;
+use Psr\Http\Message\StreamInterface;
+
+/**
+ * Implementation of PSR HTTP streams
+ */
+class CallbackStream implements StreamInterface
+{
+    /**
+     * @var callable|null
+     */
+    protected $callback;
+
+    /**
+     * @param callable $callback
+     * @throws InvalidArgumentException
+     */
+    public function __construct(callable $callback)
+    {
+        $this->attach($callback);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


src/CallbackStream.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        $this->callback = $callback;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSize()
+    {
+    }
```
weierophinney: This is specified as MUST return an integer. Considering we just added support to the `SapiEmitter` to auto-emit a Content-Length header based on the return value of this method, it's even more critical.

This would likely require that you return the strlen() of the return of getContents() — which would mean you'll likely want to have that method cache the results of calling the callback.

Doing so would also allow you to do things like read(), tell(), etc, as these could call `getContents()` if the content is not yet cached, and then start operating on the cached value. That would also allow better flexibility with emitters, as they would not need to typehint against this stream type to determine how to send the stream.

weierophinney:
In addition to my above comments, please run `./vendor/bin/phpcbf` locally, and commit the results; you have CS errors, as reported here: https://travis-ci.org/zendframework/zend-diactoros/jobs/77199455#L175

src/CallbackStream.php
oscarotero: The psr interface says:
```
/**
 * Get the size of the stream if known.
 *
 * @return int|null Returns the size in bytes if known, or null if unknown.
 */
```
So, it should be perfectly valid to return null. IMHO, this stream should not behave exactly like the standard stream (like you're proposing), because its goal is different. If I'll want to do what you say, there's no need of this class, because it can be done more easily:
```php
$result = $callable();
$response->getBody()->write($result);
```
The purpose of the callbackStream is to provide a way to send the response content in real time, so, you don't know the size or the time it will take. For example:
```php
$body = new CallbackStream(function () {
    echo 'The process is starting. Please, wait...';
    sleep(2);
    echo 'Running the process. Be patient, please';
    sleep(8);
    echo 'The process finished successfully';
});
```

PR #96, opened 2015-10-14
Fix to make withPort fully compliant with PSR-7

Nyorai:
withPort should allow null port

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -306,21 +306,23 @@ class Uri implements UriInterface
      */
     public function withPort($port)
     {
-        if (! is_numeric($port)) {
+        if (! is_numeric($port) && $port !== null) {
             throw new InvalidArgumentException(sprintf(
-                'Invalid port "%s" specified; must be an integer or integer string',
+                'Invalid port "%s" specified; must be an integer, an integer string or null',
                 (is_object($port) ? get_class($port) : gettype($port))
             ));
         }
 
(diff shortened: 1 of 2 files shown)


src/Uri.php
```diff
@@ -306,21 +306,23 @@ public function withHost($host)
...
+            $port = (int) $port;
+        }
 
         if ($port === $this->port) {
             // Do nothing if no change was made.
             return clone $this;
         }
 
-        if ($port < 1 || $port > 65535) {
+        if ($port !== null && $port < 1 || $port > 65535) {
```
Maks3w: Are you sure $port may be null here? Previously you cast to integer

Maks3w:
Missing test

test/UriTest.php
```diff
@@ -88,7 +89,7 @@ public function testWithPortReturnsNewInstanceWithProvidedPort($port)
         $this->assertNotSame($uri, $new);
         $this->assertEquals($port, $new->getPort());
         $this->assertEquals(
-            sprintf('https://user:pass@local.example.com:%d/foo?bar=baz#quz', $port),
+            sprintf('https://user:pass@local.example.com%s/foo?bar=baz#quz', $port === null ? '' : ':'.$port),
```
Maks3w: Its preferible provide the expected values as part of the data provider. What test verify this branch of logic test as expected?

PR #97, opened 2015-10-15
Add files to .gitattributes

gsouf:
Some files were not necessary. Especially changelog.md that can grow fast

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,3 +6,5 @@
 .scrutinizer.yml export-ignore
 .travis.yml export-ignore
 phpunit.xml.dist export-ignore
+CONTRIBUTING.md export-ignore
+phpcs.xml export-ignore


weierophinney:
I disagree about the CHANGELOG and the README; the former can be quite useful when upgrading, if you do not understand why code is no longer working or suddenly throwing an exception. It's far easier to look in there before going to the website to determine what changed, or, more likely, blindly filing an issue without understanding the change was intentional.

Similarly, the README file often contains information such as URIs to the issue tracker.

weierophinney:
I'll merge, but remove the README and CHANGELOG from the changeset when doing so.

Maks3w:
Only phpcs could be excluded, rest are expected to be distributable

Maks3w:
Also I prefer don't format the lines. Too noisy

weierophinney:
@Maks3w we can likely remove the CONTRIBUTING.md from the distribution packages. But the README and CHANGELOG are definitely necessary.

weierophinney:
I'll wait to merge for @gsouf to update.

Maks3w:
I would like to keep CONTRIBUTING. May help to find new contributors.

gsouf:
I understand and I accept the changes.

But a short a explanation of my opinion: 

These item (including readme and changelog) are in my opinion not useful in production.
By removing these items we can reducde the repo size by maybe 30kB. That means that creating a phar while including the whole ``vendor`` directory will be 30kB smaller.

But that feels ok, in 2015 30kB is not a big deal!

PR #98, opened 2015-10-15
Use unescaped slashes in encoded JSON by default

weierophinney:
Updated the default `$encodingOptions` to also include `JSON_UNESCAPED_SLASHES`, which is a sane default, and still complies with RFC 4627 (forward slash, or the solidus, does not need to be escaped).

### DIFF
--- a/src/Response/JsonResponse.php
+++ b/src/Response/JsonResponse.php
@@ -25,6 +25,17 @@ class JsonResponse extends Response
     use InjectContentTypeTrait;
 
     /**
+     * Default flags for json_encode; value of:
+     *
+     * <code>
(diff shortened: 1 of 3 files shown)


src/Response/JsonResponse.php
```diff
@@ -41,7 +42,7 @@ class JsonResponse extends Response
      * @param int $encodingOptions JSON encoding options to use.
      * @throws InvalidArgumentException if unable to encode the $data to JSON.
      */
-    public function __construct($data, $status = 200, array $headers = [], $encodingOptions = 15)
+    public function __construct($data, $status = 200, array $headers = [], $encodingOptions = 79)
```
Ocramius: That last hardcoded number bothers me a lot. Can it be moved to a class constant?

weierophinney:
@Ocramius done. :)

Issue #100, opened 2015-10-26
MessageTrait::withHeader does not check for existing header names


pdscopes:
Not a massive issue but when when calling ```withHeader``` there is no check for an existing header name inside ```$this->headers```. This means that the newly Message created message could have a duplicated property. This can only occur when the added header name has a different capitalisation than the already stored header of the same name. i.e.:

```
// Assume there is a response message called $response
// Also assume that $response has a header set 'Content-Type' => ['application/json']
$new = $response->withHeader('Content-type', ['text/html']);
```

```$new``` will have one value in the ```$this->headerNames``` map but 'Content-Type' and 'Content-type' in ```$this->headers```.

A potential solution is to perform a ```$this->hasHeader($name)``` check and unsetting that header value if true _before_ setting the value in ```$new```.

**Comment by Maks3w** [comment]
I confirm getHeaders will return old and new headers 

Could you send a PR with the fix?

**Comment by pine3ree** [comment]
Another solution could be using the normalized header name for the $headers property of  MessageTrait but it would require many more changes: 
- withHeader(...)
```php
public function withHeader($name, $value)
{
    //...
    $normalized = strtolower($name);

    $new = clone $this;
    $new->headerNames[$normalized] = $name;
    $new->headers[$normalized]     = $value;

    return $new;
}
```
- withAddedHeader(...)
```php
public function withAddedHeader($name, $value)
{
    //...
    $normalized = strtolower($name);

    $new = clone $this;
    $new->headers[$normalized] = array_merge($this->headers[$normalized], $value);

    return $new;
}
```
- getHeaders() methods : fetch the original case $name from $headerNames in instead of simply returning the $headers property 
```php
public function getHeaders() 
{
    $headers = [];
    foreach ($this->headers as $name => $value) {
        $headers[$this->headerNames[$name]] = $value;
    }

    return $headers;
}
```
- getHeader(...) methods: here it won't be needed to fetch the original header name anymore
```php
public function getHeader($name)
{
    if (! $this->hasHeader($name)) {
        return [];
    }

    $value  = $this->headers[strtolower($name)];
    $value  = is_array($value) ? $value : [$value];

    return $value;
}
```

... and so on...
[...]

PR #101, opened 2015-10-26
Fixing issue #100 withHeader does not remove old header value

pdscopes:
Pull request to fix Issue #100

### DIFF
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -204,7 +204,7 @@ trait MessageTrait
         $normalized = strtolower($header);
 
         $new = clone $this;
-        if($new->hasHeader($header)) {
+        if ($new->hasHeader($header)) {
             unset($new->headers[$new->headerNames[$normalized]]);
         }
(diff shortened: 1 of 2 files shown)


Maks3w:
Need a test case

src/MessageTrait.php
```diff
@@ -204,6 +204,9 @@ public function withHeader($header, $value)
         $normalized = strtolower($header);
 
         $new = clone $this;
+        if($new->hasHeader($header)) {
+            unset($new->headers[$new->headerNames[$normalized]]);
```
Maks3w: I wonder if this could be done always and remove the if

pdscopes: The problem there is, if  ```$new->headerNames[$normalized]```doesn't exist PHP will throw a warning.
I guess we could not use ```$new->hasHeader()``` and directly call ```array_key_exists($normilized, $new->headerNames)```.
What do you think?

PR #102, opened 2015-10-27
Clone support

eguvenc:
If some one intend to "clone $request" object uri class should be clonable otherwise request object cannot clone completely. This feature also allow to us create our HMVC design pattern.

### DIFF
--- a/src/RequestTrait.php
+++ b/src/RequestTrait.php
@@ -90,6 +90,17 @@ trait RequestTrait
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Ocramius:
@eguvenc requires testing

Also, a `Uri` is immutable, so cloning it shouldn't be necessary.

eguvenc:
Sorry this is my mistake.
before i was use clone method instead of creating new instance, after that creating new instance of $request object  fixed the issue.

PR #103, opened 2015-11-06
Simplify status code

snapshotpl:
I decided to simplify set status code. Why? Constructor's phpdocs says that it's expected `$status` variable as int, not null, only int. I know that make BC, however it should be done in my opinion

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -102,7 +102,7 @@ class Response implements ResponseInterface
     /**
      * @var int
      */
-    private $statusCode = 200;
+    private $statusCode;
 
     /**
      * @param string|resource|StreamInterface $stream Stream identifier and/or actual stream resource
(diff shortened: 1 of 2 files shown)


test/ResponseTest.php
```diff
@@ -151,10 +151,18 @@ public function testConstructorIgonoresInvalidHeaders()
             'x-valid-string' => [ 'VALID' ],
             'x-valid-array' => [ 'VALID' ],
         ];
-        $response = new Response('php://memory', null, $headers);
+        $response = new Response('php://memory', 200, $headers);
         $this->assertEquals($expected, $response->getHeaders());
     }
 
+    /**
+     * @expectedException \InvalidArgumentException
```
Maks3w: Please follow the same convention as the rest of tests. Use `setExpectedException` and fill the second argument.

src/Response.php
```diff
@@ -159,9 +156,8 @@ public function getReasonPhrase()
      */
     public function withStatus($code, $reasonPhrase = '')
     {
-        $this->validateStatus($code);
```
Maks3w: Revert all changes after this line. There is a clear intention of not provide any kind of setter.

src/Response.php
```diff
@@ -120,12 +120,9 @@ public function __construct($body = 'php://memory', $status = 200, array $header
             );
         }
 
-        if (null !== $status) {
-            $this->validateStatus($status);
```
Maks3w: Restore the call to this method.

weierophinney:
What purpose does a null status code serve? I honestly do not understand the rationale behind this change.

Maks3w:
What he say is the argument should be always integer however the actual code implemented allow null

src/Response.php
snapshotpl: Ok, but why? We need to repeat this code hear and in contructor, so private setter keep setting status code always correct

PR #104, opened 2015-11-06
Change variable name to be complain with phpdoc

snapshotpl:


.gitignore
```diff
@@ -2,3 +2,4 @@ doc/html/
 vendor/
 composer.lock
 phpunit.xml
+/nbproject/private/
```
samsonasik: it should be reverted

Maks3w:
1. Body is a better variable name. You don't name a variable "string" neither "integer" you usually name it by his content not by his type.

2. Headers was well assigned, save memory and unsafe headers (unfiltered) should not be used ever.

3. Modify gitignore file is totally out of the scope.


I'll close this PR because has too many faults. I suggest you open a new one modifying the PHPDoc tag.

PR #99, opened 2015-12-15
Header value not string

weierophinney:
I have tried to create a request and I added header:
```php
[
  'X-User-Id': 123,
]
```
After this my request doesn't have `X-User-Id` header. After debug I have found this https://github.com/zendframework/zend-diactoros/blob/1037ed1cb4e240eba665597e4c373e57fb9ea24c/src/MessageTrait.php#L341

Should we throw exception instead of continue code execution?

### DIFF
--- a/src/MessageTrait.php
+++ b/src/MessageTrait.php
@@ -352,7 +352,7 @@ trait MessageTrait
 
             if (is_array($value)) {
                 array_walk($value, function ($item) {
-                    if (! is_array($item) && ! is_string($item) && ! is_numeric($item)) {
+                    if (! is_string($item) && ! is_numeric($item)) {
                         throw new InvalidArgumentException(sprintf(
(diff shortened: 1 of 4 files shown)


weierophinney:
I've converted this issue into a pull request, and it contains a patch implementing the changes. These include:

- Numeric (integer *and* float) values are allowed for header values.
- Exceptions are raised when:
  - The header name is invalid
  - The header value is invalid
  - Any item in a header value array is invalid

snapshotpl:
@weierophinney thanks for fix that. However I see that fix introduce code-duplication. Let's see at:
https://github.com/zendframework/zend-diactoros/blob/master/src/MessageTrait.php#L239
https://github.com/zendframework/zend-diactoros/blob/master/src/MessageTrait.php#L195
https://github.com/zendframework/zend-diactoros/blob/master/src/MessageTrait.php#L339

Issue #89, opened 2015-08-21
Serialization of Request


dzentota:
```php

    public function testSerialization()
    {
        $request = new \Zend\Diactoros\Request('https://example.com', 'GET');
        $serialized = \Zend\Diactoros\Request\Serializer::toString($request);
        $unserialized = \Zend\Diactoros\Request\Serializer::fromString($serialized);
        $this->assertEquals((string)$request->getUri(), (string)$unserialized->getUri());
        $this->assertEquals($request->getHeaders(), $unserialized->getHeaders());
        $this->assertEquals($request->getRequestTarget(), $unserialized->getRequestTarget());
    }
```

```
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'https://example.com'
+'/'

```

Absolute-form of request-target should be used during serialization of Request object because otherwise we lose information about uri schema (https in this case). 
```php
 public function testSerialization()
    {
        $request = new \Zend\Diactoros\Request();
        $serialized = \Zend\Diactoros\Request\Serializer::toString($request);
        $unserialized = \Zend\Diactoros\Request\Serializer::fromString($serialized);
        $this->assertEquals((string)$request->getUri(), (string)$unserialized->getUri());
        $this->assertEquals($request->getHeaders(), $unserialized->getHeaders());
        $this->assertEquals($request->getRequestTarget(), $unserialized->getRequestTarget());
    }
```

```
UnexpectedValueException: Invalid request line detected
```
Maybe there should be another approach for serializing of Request which allows serialization of partial filled objects?

dzentota:
Possible fix is to rewrite getRequestTarget() as:
```php
    public function getRequestTarget()
    {
        if (null !== $this->requestTarget) {
            return $this->requestTarget;
        }

        if (! $this->uri) {
            return '/';
        }
        return (string) $this->uri;
    }

```

weierophinney:
I get your point. However, the specification for `getRequestTarget()` is such that it will return the origin-form **always**, unless you have specifically set the request target using `withRequestTarget()`. In your example, you did not do so, which is equivalent to the following message:

```http
GET / HTTP/1.1
Host: example.com
```

The request target is of interest only when actually making a request, and is used to make the connection in the first place. As such, I'd argue the current behavior is correct; otherwise, we end up with absolute-form by default, which is, frankly, not used often, and unexpected in most instances.

[...]

PR #112, opened 2015-12-16
Allows deserialization of serialized requests with omitted http method

dzentota:


weierophinney:
Per my comment on #89, I'm not sure this makes sense. Lack of an HTTP method means that the request is in an invalid state, and de/serialization of invalid messages probably *should not* be supported, as it makes parsing impossible.

dzentota:
Closed in favor of https://github.com/zendframework/zend-diactoros/pull/117

PR #115, opened 2015-12-16
Added stream rewind in JsonResponse constructor

NikolaySl:


weierophinney:
Thanks for reporting this and providing a patch! I was about to merge, and realized the issue was also present in the HTML and text response types! I'm updating those as part of the merge.

NikolaySl:
Thanks for very fast fixes. 

When new version will be released?

weierophinney:
@NikolaySl Already done; 1.3.1 has the fixes.

mnapoli:
@weierophinney how are we supposed to write to the stream afterwards since it has been rewinded?

Let's say we have a middleware that adds additional content to the stream, `$response->getBody()->write()` will overwrite existing data in the stream?

mnapoli:
Moved my question to #195, it affects all response classes.

Issue #120, opened 2015-12-22
Protocol version in $_SERVER isn't parsed


akrabat:
`$_SERVER['PROTOCOL_VERSION']` is not parsed in `ServerRequestFactory` and hence we don't set the `protocolVersion` property to the correct value.

mtymek:
Here you go @akrabat! #122 :)

weierophinney:
Fixed with #122

PR #122, opened 2015-12-22
Parse protocol version

mtymek:
As requested in #120.

### DIFF
--- a/src/ServerRequestFactory.php
+++ b/src/ServerRequestFactory.php
@@ -14,6 +14,7 @@ use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\MessageInterface;
 use Psr\Http\Message\UploadedFileInterface;
 use stdClass;
+use UnexpectedValueException;
 
 /**
  * Class for marshaling a request object from the current PHP environment.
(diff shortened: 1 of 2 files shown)


akrabat:
Looks good to me.

src/ServerRequestFactory.php
```diff
@@ -70,6 +71,7 @@ public static function fromGlobals(
         );
 
         return $request
+            ->withProtocolVersion(static::marshalProtocolVersion($_SERVER))
```
snapshotpl: I think that you should use here `$server`

src/ServerRequestFactory.php
```diff
@@ -70,6 +71,7 @@ public static function fromGlobals(
...
             ->withCookieParams($cookies ?: $_COOKIE)
```
snapshotpl: I'm little worried about this. Why we need to create 5 instances of `ServerRequest`? `ServerRequest` should be able to be constructed with all possible args. Maybe we can named constructor here?

snapshotpl: I prepared patch for that #124

src/ServerRequestFactory.php
weierophinney: Agreed; I'll make that change on merge.

src/ServerRequestFactory.php
```diff
@@ -455,4 +457,23 @@ private static function normalizeNestedFileSpec(array $files = [])
...
         return $normalizedFiles;
     }
+
+    /**
+     * Return HTTP protocol version (X.Y)
+     *
+     * @param array $server
+     * @return string
+     */
+    public static function marshalProtocolVersion(array $server)
```
weierophinney: This should also be private, for consistency with the other methods. I'll make that change on merge.

PR #124, opened 2015-12-22
Server request can be created with all possible params

snapshotpl:


weierophinney:
Not entirely happy with this, as that's a TON of arguments. That said, it does prevent extraneous instances. (This is a micro-optimization, mind you; those instances will be GC'd almost immediately, as there will be no lingering references.)

Issue #121, opened 2015-12-22
ServerRequest and fragments


oscarotero:
Hello. I'm getting this error:

```
Invalid path provided; must not contain a URI fragment at /var/www/asociaciondag.org/releases/20151222083807/vendor/zendframework/zend-diactoros/src/Uri.php:356)
```

The `$_SERVER` global variable has the following values:

```
{"REDIRECT_HTTPS":"on","REDIRECT_SSL_TLS_SNI":"dag.gal","REDIRECT_STATUS":"200","HTTPS":"on","SSL_TLS_SNI":"dag.gal","HTTP_USER_AGENT":"WordPress/4.3.1; http://www.arinspunk.gal","HTTP_HOST":"dag.gal","HTTP_ACCEPT":"*/*","HTTP_ACCEPT_ENCODING":"deflate;q=1.0, compress;q=0.5, gzip;q=0.5","SERVER_SIGNATURE":"<address>Apache/2.4.7 (Ubuntu) Server at dag.gal Port 443</address>\n","SERVER_SOFTWARE":"Apache/2.4.7 (Ubuntu)","SERVER_NAME":"dag.gal","SERVER_ADDR":"178.62.79.169","SERVER_PORT":"443","REMOTE_ADDR":"188.121.41.219","DOCUMENT_ROOT":"/var/www/asociaciondag.org/current/public","REQUEST_SCHEME":"https","CONTEXT_PREFIX":"","CONTEXT_DOCUMENT_ROOT":"/var/www/asociaciondag.org/current/public","SCRIPT_FILENAME":"/var/www/asociaciondag.org/current/public/index.php","REMOTE_PORT":"49427","REDIRECT_URL":"/es/feed2015","GATEWAY_INTERFACE":"CGI/1.1","SERVER_PROTOCOL":"HTTP/1.0","REQUEST_METHOD":"HEAD","QUERY_STRING":"","REQUEST_URI":"/es/feed2015#oscar-otero","SCRIPT_NAME":"/index.php","PHP_SELF":"/index.php","REQUEST_TIME_FLOAT":1450798110.232,"REQUEST_TIME":1450798110}
```

Note that the `REQUEST_URI` is "/es/feed2015#oscar-otero" (with fragment) so I guess this makes to fail `serverRequestFactory` because the path is handled with the fragment.

oscarotero:
Hi.
I've created a PR (#132) to fix this error. :smiley:

to-masz:
[...]

PR #132, opened 2015-12-30
Handle uri fragments in the serverRequestFactory

oscarotero:
This fixes the issue #121


weierophinney:
Just a quick note: typically the fragment is not included in the HTTP request line, which makes this mostly pointless. I'm merging it just to be complete.

PR #117, opened 2015-12-17
Raise exception on serialization of Request object with emtpy HTTP me…

dzentota:
…thod

PR #136, opened 2016-01-05
Correction of regular expression for HTTP 2.0

ziege:
The "SERVER_PROTOCOL" value returned by Apache is "HTTP/2" (and not "HTTP/2.0"). I adjusted the regular expression to match it.

### DIFF
--- a/src/ServerRequestFactory.php
+++ b/src/ServerRequestFactory.php
@@ -465,7 +465,7 @@ abstract class ServerRequestFactory
             return '1.1';
         }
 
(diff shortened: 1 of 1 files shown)


mtymek:
Your regexp works fine, but it would be nice to add a unit test.

xtreamwayz:
I've got the unit test. If you want I can create a PR for it.

https://github.com/xtreamwayz/zend-diactoros/blob/hotfix/marshal-http2-protocol/test/ServerRequestFactoryTest.php#L458-L476

weierophinney:
Fixed already when merging #117.

PR #138, opened 2016-01-08
Fix minor typo in contributing guide

settermjd:
I found a minor typo in the contributing guide wrt the diactoros repository.


samsonasik:
no, it was already right.

settermjd:
Really? It didn't work for me. I'll check again though.

weierophinney:
@samsonasik The change from @settermjd is correct; when using the git protocol, the URI follows normal URI semantics. It's when you use the SSH variant that you swap the `/` for a `:` following the domain.

PR #139, opened 2016-01-22
fix callback call

n0o0bSina:


src/CallbackStream.php
```diff
@@ -155,7 +155,7 @@ public function getContents()
     {
         $callback = $this->detach();
 
-        return $callback ? $callback() : '';
+        return $callback ? call_user_fanc($callback) : '';
```
sasezaki: `fanc` ?
https://travis-ci.org/zendframework/zend-diactoros/jobs/104034138#L163
you should add test to determine what you add .
https://github.com/zendframework/zend-diactoros/blob/master/test/CallbackStreamTest.php

This PR's clear title may be ..
CallbackStream should be allow all of callable, current implentation is allowed only Closure .

weierophinney:
I'm merging, but reverting the change.

In my local merge branch, I've added tests for arbitrary PHP callbacks. Starting with PHP 5.4, you *can* call any PHP callback using the `$callback()` notation, making the change unnecessary.

PR #140, opened 2016-01-24
Remove unnecessary test in URI constructor

krzysiekpiasecki:
In 99,9% we should create object using constructor and parse uri argument

- lighter constructor
- lower complexity

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -99,10 +99,8 @@ class Uri implements UriInterface
                 (is_object($uri) ? get_class($uri) : gettype($uri))
             ));
         }
-
-        if (! empty($uri)) {
-            $this->parseUri($uri);
-        }
+        
+        $this->parseUri($uri);
     }
 
     /**


Ocramius:
The `Uri` object is used massively, and therefore this is a minor performance optimization that was introduced on purpose. 

Closing as `invalid`

krzysiekpiasecki:
>> The  Uri  object is used massively

Hello,

What does it mean? 'One' URI per milion users or milion users with 'one' URI? The difference per one object is ~0.001 μ.

But it's not a clue. IMHO you wrong:
- This patch doesn't affect PSR-7 interface because of cloning,
- The efficiency is higher,
- The complexity is lower,
- The constructor is lighter.

Take a look  again into constructor:
```php
        if (! empty($uri)) {
            $this->parseUri($uri);
        }
```

Theoretically one call less. But when, when URI is empty? Developers build URI with argument, so constructor MUST call the parseURI.


I don't that above is a mature optimization.

PR #142, opened 2016-01-27
Verbose exception message from header validators

mtymek:
With current version of Diactoros I often see exceptions like "Invalid header value". It doesn't show which header is not valid, making it difficult to see what's the problem. 
With this PR server logs will clearly show which header caused exception.

### DIFF
--- a/src/HeaderSecurity.php
+++ b/src/HeaderSecurity.php
@@ -129,7 +129,7 @@ final class HeaderSecurity
     public static function assertValid($value)
     {
         if (! self::isValid($value)) {
-            throw new InvalidArgumentException('Invalid header value');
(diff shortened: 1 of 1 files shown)


src/HeaderSecurity.php
```diff
@@ -129,7 +129,7 @@ public static function isValid($value)
     public static function assertValid($value)
     {
         if (! self::isValid($value)) {
-            throw new InvalidArgumentException('Invalid header value');
+            throw new InvalidArgumentException("'$value' is not valid header value");
```
weierophinney: We typically use `sprintf()` for interpolation in exception messages; I'll make that change on merge.

PR #157, opened 2016-03-12
Small simplification

nazar-pc:
Original code seems over-engineered

### DIFF
--- a/src/ServerRequestFactory.php
+++ b/src/ServerRequestFactory.php
@@ -198,21 +198,18 @@ abstract class ServerRequestFactory
     {
         $headers = [];
         foreach ($server as $key => $value) {
-            if ($value && strpos($key, 'HTTP_') === 0) {
-                $name = strtr(substr($key, 5), '_', ' ');
-                $name = strtr(ucwords(strtolower($name)), ' ', '-');
-                $name = strtolower($name);
+            if ($value) {
+                if (strpos($key, 'HTTP_') === 0) {
+                    $name = substr($key, 5);
+                } elseif (strpos($key, 'CONTENT_') === 0) {
+                    $name = $key;
+                } else {
+                    continue;
+                }
 
-                $headers[$name] = $value;
-                continue;
-            }
-
-            if ($value && strpos($key, 'CONTENT_') === 0) {
-                $name = substr($key, 8); // Content-
-                $name = 'Content-' . (($name == 'MD5') ? $name : ucfirst(strtolower($name)));
+                $name = strtr($name, '_', '-');
                 $name = strtolower($name);
                 $headers[$name] = $value;
-                continue;
             }
         }
 


src/ServerRequestFactory.php
```diff
@@ -198,21 +198,18 @@ public static function marshalHeaders(array $server)
...
+                $name = strtr($name, '_', '-');
                 $name = strtolower($name);
-
                 $headers[$name] = $value;
-                continue;
-            }
-
-            if ($value && strpos($key, 'CONTENT_') === 0) {
-                $name = substr($key, 8); // Content-
-                $name = 'Content-' . (($name == 'MD5') ? $name : ucfirst(strtolower($name)));
```
Ocramius: Was this just dropped? Was there no coverage for it?

src/ServerRequestFactory.php
```diff
@@ -198,21 +198,18 @@ public static function marshalHeaders(array $server)
     {
         $headers = [];
         foreach ($server as $key => $value) {
-            if ($value && strpos($key, 'HTTP_') === 0) {
-                $name = strtr(substr($key, 5), '_', ' ');
-                $name = strtr(ucwords(strtolower($name)), ' ', '-');
+            if ($value) {
```
Ocramius: This code has a much higher cyclomatic complexity, and is also harder to understand, IMO

src/ServerRequestFactory.php
nazar-pc: If you analyze it correctly - it basically does nothing.
First we cut `Content-` at the beginning, then add if unconditionally (why?). After this we check whether remainder is equal to `MD5` and if so - pass it directly, otherwise we make reminder lowercase and than make first character uppercase. After all manipulations we make it lowercase (which basically neglects everything from previous sentence).
So:
* cutting `CONTENT_` and replacing with `Content-` seems reasonable, but can be just replaced with `strtr()` (the real difference in practice is just one symbol) in order to keep is homogeneous with previous block of code
* all the game with repeated upper/lowercasing doesn't make any sense

As far as I saw there are test cases for all situations touched by these changes.

src/ServerRequestFactory.php
nazar-pc: I read it in following way:
* we are only interested if there is some value
* then we are curious what key we have
 * if key starts with `HTTP_` - drop prefix
 * if key starts with `CONTENT_` - take it as is
 * ignore key otherwise
* normalize key (common for both cases for simplicity)
 * replace `_` with `-` in key
 * make key lowercase

Added cyclomatic complexity is only because we've decoupled common operations for `HTTP_*` and `CONTENT_*` keys, otherwise it is the same. I thought there is no reason to duplicate 3 exactly the same lines in 2 neighbor blocks.

weierophinney:
I'm going to keep the code as-is, for the following reasons:

- Having two separate blocks makes it quite clear what the two header sources are within `$_SERVER`. I personally feel this is important for developers trying to understand how `$_SERVER` works within the standard SAPI implementations.
- Lower cyclomatic complexity.

nazar-pc:
I can refactor it to have 2 separate blocks, but leaving as-is means by having lower cyclomatic complexity there will be much more CPU cycles wasted for nothing

weierophinney:
> there will be much more CPU cycles wasted for nothing

This code runs at the start of a request, and will be executed for only as many entries as are present in `$_SERVER`. The changes are a micro-optimization at best, and sacrifice clarity.

nazar-pc:
Final change from my side, refactored as promised: https://github.com/nazar-pc/zend-diactoros/commit/c3cc4b42fcc83348e638006376313fe73b97e42c
If you really really think that clarity was sacrificed, then I have nothing to add.

Issue #17, opened 2015-05-25
Rename to something that will help people know what the hell this is


GeeH:
I love clever and unusual names as much as the next man, but this name is just horrid.

GeeH:
The automatic wontfix label makes me :cry:

weierophinney:
Unless you have suggestions, we're not likely to change the name at this point. We couldn't may name it "http" for obvious reasons, "HttpMessage" would never return relevant results, and "Hermes" or "Mercury" are already flooded with results. We briefly considered "zttp" or "zhttp", but ultimately the them out.

The criteria we have:

- cannot conflict with existing ZF component names
- cannot conflict with existing php project/company names
- should have some affinity with either http or messaging
- should be unique enough to rise to the top of search results

I'll gladly consider names that fulfill those criteria, but the name will need to be quite compelling to trigger a rename at this point.

GeeH:
I get that - I'm actually not trying to be argumentative (for once). Personally I just feel that a non-unique descriptive name trumps Googlability.

weierophinney:
It's not just about search; part of this is about being able to track reach. As an example, ZF often appears to trend lower than other frameworks, while simultaneously having more downloads. The reason is interesting: two words in the name, one common, and the other a brand shared amongst several products skews to lower search ranking, which makes understanding our reach difficult. 

Hence, the argument for something more unique. :-)

GeeH:
I get that, it feel so pretentious, I just hate this name.

mtymek:
Agree with GeeH, at the beginning I kept forgetting how it is spelled :) 
Difficult name won't help with getting more reach.

GeeH:
Why can't it just be Zend\Http 3.0?

Maks3w:
`zend-psr7`

akrabat:
Zend\Http clearly doesn't meet the requirements set out above.

GeeH:
Obviously, but my question is why Zend\Http 3.x can't be the PSR-7 implementation? Surely this is exactly the kind of thing that the repositories were split for?

weierophinney:
> why Zend\Http 3.x can't be the PSR-7 implementation? Surely this is exactly the kind of thing that the repositories were split for?

Yes and no.

In this case, the architecture of `Zend\Http` is radically different than a PSR-7 implementation:

- it provides a client implementation
- it provides concrete header implementations
- it provides a cookie-jar implementation

[...]

PR #160, opened 2016-03-17
Accept HTTP/2 protocol

xtreamwayz:
This is still failing on version 1.3.4 even though it should have been fixed with #117?


weierophinney:
Separate issues. #17 was addressing general protocol validation in the request object (as set in the constructor or via a `withProtocolVersion()` call), while your patch is looking at how the `ServerRequestFactory` parses it from the incoming server params.

PR #162, opened 2016-03-19
Fix invalid default method

snapshotpl:
Currently, when method is not provided it's set as empty string. It's invalid state, because it's not possible to set new method (using `withMethod`) as empty string. It's also other possible invalid state - we can set new method as `null`, which is not compatible with `RequestInterface`.

### DIFF
--- a/src/Request/Serializer.php
+++ b/src/Request/Serializer.php
@@ -77,9 +77,6 @@ final class Serializer extends AbstractSerializer
     public static function toString(RequestInterface $request)
     {
         $httpMethod = $request->getMethod();
-        if (empty($httpMethod)) {
-            throw new UnexpectedValueException('Object can not be serialized because HTTP method is empty');
-        }
         $headers = self::serializeHeaders($request->getHeaders());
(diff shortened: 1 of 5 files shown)


weierophinney:
This is a slight BC break in that `toString()` can now raise an exception. As such, I'm going to push it to the next minor release.

weierophinney:
Hi, @snapshotpl!

I've been reviewing this in a bit more detail, and I've decided to push it to a future 2.0 release.

The reason I'm pushing it to a new major release is because it breaks existing assumptions codified in the tests, particularly around the fact that we previously allowed a client-side request to have a nullable request method; changing this assumption, while a net-positive, causes potential breakage for existing users who may have depended on this "feature".

(Interestingly, PSR-7 does not indicate that there MUST be a value present for the HTTP method, only that `getMethod()` must return a string. The current implementation does that, except in one case — we currently allow `null`, which we should not. More generally, PSR-7 does not dictate that a given instance MUST have a valid state, and there are a variety of ways to create an instance with invalid state, including omitting the URI!)

Issue #155, opened 2016-03-11
Uri implementation does not support the utf8 uri's


zstate:
Uri implementation does not support utf8 uri's because of PHP parse_url function (see https://secure.php.net/manual/en/function.parse-url.php#114817)

zf2timo:
Can you create a Unit Test which demonstrates the issue?

zstate:
```
class UriTest extends PHPUnit_Framework_TestCase
{
    /**
     * This test fails, but it should not
     */
    public function testUtf8Uri()
    {
        $uri = new \Zend\Diactoros\Uri('http://ουτοπία.δπθ.gr/');

        $this->assertEquals('ουτοπία.δπθ.gr' , $uri->getHost());
    }
}

```

zf2timo:
I added your provided Test locally, and the Test did not failed.
Does it fail on your local System?

And please also created a PR with this Unit Test, so we can see, the result on Travis.

zstate:
Yeah looks like it is my environment which is Mac OS X EL Capitan 10.11.2 (15C50) with PHP 5.5.30 (cli) (built: Oct 23 2015 17:21:45). 

When I run the tests I'm getting this output:

There was 1 failure:

1) ZendTest\Diactoros\UriTest::testUtf8Uri
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'ουτοπία.δπθ.gr'
+'ο?_?_ο?_ία.δ?_θ.gr'


Not sure what's going on and what I need to configure to make it work. Running tests on CentOS works fine and the travis build successful.

zf2timo:
I think this issue can be closed for now.
As seen in PR #159 the Travis Build is successful.

Pudge601:
I've run into this issue myself when trying to run the unit tests on my machine, and it appears to be due to locale settings, specifically the `LC_CTYPE` ([docs](http://php.net/manual/en/function.setlocale.php)) locale setting.

This locale setting is automatically determined based on the `LANG` environment variable, which in my case was set to `en_GB.UTF-8`, and the output of the unit test with this environment was;

```
1) ZendTest\Diactoros\UriTest::testUtf8Uri
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'ουτοπία.δπθ.gr'
+'��_�_��_��.��_�.gr'
```

[...]

PR #166, opened 2016-03-29
Added support for utf8 path

agoosev:
Currently, we can't work with non-latin letters in path. For example: http://example.com/тестовый_путь/ or http://example.com/flüg/. I resolve this problem by adding in `CHAR_UNRESERVED ` constant `\pL` for support utf-8 letters.

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -36,7 +36,7 @@ class Uri implements UriInterface
      *
      * @const string
      */
-    const CHAR_UNRESERVED = 'a-zA-Z0-9_\-\.~';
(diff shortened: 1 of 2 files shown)


zf2timo:
Looks similar to #155 

An Unit Test with UTF-8 Characters was already added in #159.

agoosev:
zf2timo, yes, but #155 covers only cases with utf8 domains. In this pull request I covered cases with utf8 path of url.

zf2timo:
@agoosev you are right.

weierophinney:
Thanks, @agoosev\!

PR #169, opened 2016-04-18
Fix reponse serialization if response body is empty.

jbaron-gingco:
The newline after the headers was missing if the response body was empty.


weierophinney:
Thanks, @jbaron-gingco

PR #173, opened 2016-05-29
marshall nicely formatted header names

pine3ree:
otherwise we could just rewrite the method simply as:
```php
    public static function marshalHeaders(array $server)
    {
        $headers = [];
                $name = strtr(strtolower(substr($key, 5)), '_', '- ');
                $headers[$name] = $value;
                continue;
            }

            if ($value && strpos($key, 'CONTENT_') === 0) {
                $name = 'content-' . strtolower(substr($key, 8));
                $headers[$name] = $value;
                continue;
            }
        }

        return $headers;
    }
```

without ucwords/ucfirst (redundant because overwritten by last strtolower) calls

header array created via marshallHeader are filtered via `MessageTrait::filterHeaders(array $originalHeaders)`. Original/well-formatted headers are kept in the $header property, lowercased header names are used as keys of $headerNames.

### DIFF
--- a/src/ServerRequestFactory.php
+++ b/src/ServerRequestFactory.php
@@ -199,7 +199,6 @@ abstract class ServerRequestFactory
             if ($value && strpos($key, 'HTTP_') === 0) {
                 $name = strtr(substr($key, 5), '_', ' ');
                 $name = strtr(ucwords(strtolower($name)), ' ', '-');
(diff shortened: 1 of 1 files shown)


pine3ree:
oops, i saw the test assertions....i will add a different PR to avoid redundant operations then....

PR #174, opened 2016-05-30
remove overwritten function calls

pine3ree:
at the end HTTP_ and CONTENT_ header names from $_SERVER and mocked request are normalized to lowercase, so nice formatting function calls (ucwords/ucfirst) are redundant.

### DIFF
--- a/src/ServerRequestFactory.php
+++ b/src/ServerRequestFactory.php
@@ -197,7 +197,7 @@ abstract class ServerRequestFactory
         $headers = [];
         foreach ($server as $key => $value) {
             if ($value && strpos($key, 'HTTP_') === 0) {
-                $name = strtr(strtolower(substr($key, 5)), '_', '- ');
+                $name = strtr(strtolower(substr($key, 5)), '_', '-');
                 $headers[$name] = $value;
                 continue;
             }
--- a/src/ServerRequestFactory.php
+++ b/src/ServerRequestFactory.php
@@ -197,18 +197,13 @@ abstract class ServerRequestFactory
         $headers = [];
         foreach ($server as $key => $value) {
             if ($value && strpos($key, 'HTTP_') === 0) {
-                $name = strtr(substr($key, 5), '_', ' ');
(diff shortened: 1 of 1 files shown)


src/ServerRequestFactory.php
```diff
@@ -197,18 +197,13 @@ public static function marshalHeaders(array $server)
         $headers = [];
         foreach ($server as $key => $value) {
             if ($value && strpos($key, 'HTTP_') === 0) {
-                $name = strtr(substr($key, 5), '_', ' ');
-                $name = strtr(ucwords(strtolower($name)), ' ', '-');
-                $name = strtolower($name);
-
+                $name = strtr(strtolower(substr($key, 5)), '_', '- ');
```
snapshotpl: Remove space after `-`

snapshotpl: Missing unit test for that?

pine3ree:
@snapshotpl 

Thanks, that's what happen when working late hours on sundays.... 

- I first added a PR (https://github.com/zendframework/zend-diactoros/pull/173) to keep well-formatted marhalled header names, but then i found out that a unit test required them to be lowercased.

- I kept the extra line for clarity and to have a similar structure for the 2 special cases (HTTP_ and CONTENT_)

- About unit test: there already was a test for that. My typo still passed the test due t the nature of strtr(). If str_replace were used it wouldn't have passed.

weierophinney:
I don't understand the purpose of this change; what problem is it trying to solve?

pine3ree:
@weierophinney 
Hello,
i believe there is no point in calling ucwords `ucwords(strtolower($name))` in line 201 and 210 of the original code because just after that $name (lines 202, 211 original code) is eventually lowercased before being used as an array key.

kind regards

weierophinney:
@pine3ree Okay, makes more sense now; missed that the strtolower call was happening previously.

weierophinney:
Thanks, @pine3ree

PR #179, opened 2016-05-31
moveTo: check targetDirectory and more

pine3ree:
- test for already moved file 1st (or 2nd) : testing a boolean can avoid heavier tests
- unify string and empty string in a single test: the exception message suggests both requirements


weierophinney:
Thanks, @pine3ree

PR #188, opened 2016-07-05
Used relative stream to simplify emitBodyRange. Removes necessity to check lower bound

hannesvdvreken:


samsonasik:
need rebase

hannesvdvreken:
See #202

Issue #192, opened 2016-08-03
Unexpected symbol


undercloud:
```PHP
$uri = new Zend\Diactoros\Uri;

$uri = $uri
	->withHost('google.com');
// string(11) "google.com:" 
var_dump((string)$uri);
```

adamlundrigan:
Confirmed.

Test Case:
```php
    public function testUriDoesNotAppendColonToHostIfPortIsEmpty()
    {
        $uri = (new Uri())->withHost('google.com');

        $this->assertNotEquals('google.com:', (string)$uri);
    }
```

Result:

```
There was 1 failure:

1) ZendTest\Diactoros\UriTest::testUriDoesNotAppendColonToHostIfPortIsEmpty
Failed asserting that 'google.com:' is not equal to <string:google.com:>.

/var/www/zendframework/zend-diactoros/test/UriTest.php:576
```

PR #194, opened 2016-08-14
Update Uri.php

undercloud:
https://github.com/zendframework/zend-diactoros/issues/192


adamlundrigan:
I suggest fixing this in method `isNonStandardPort`, which will fix all instances of the problem rather than just the one you're experiencing.  Switching the order of the two if statements so the host/port is checked before the scheme should do the trick.

samsonasik:
I've created alternative fixes at #197

weierophinney:
Closing in favor of #197.

PR #196, opened 2016-08-19
Allow for UriInterface

Nyholm:
It is supported by `$this-> initialize()`

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -24,7 +24,7 @@ class Request implements RequestInterface
     use MessageTrait, RequestTrait;
 
     /**
-     * @param null|string $uri URI for the request, if any.
+     * @param null|string|UriInterface $uri URI for the request, if any.
      * @param null|string $method HTTP method for the request, if any.
      * @param string|resource|StreamInterface $body Message body, if any.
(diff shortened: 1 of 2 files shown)


Nyholm:
I see that there are some build failures but they are not related to this change.

src/Request.php
```diff
@@ -24,7 +24,7 @@ class Request implements RequestInterface
     use MessageTrait, RequestTrait;
 
     /**
-     * @param null|string $uri URI for the request, if any.
+     * @param null|string|UriInterface $uri URI for the request, if any.
```
shadowhand: As per https://github.com/Nyholm/zend-diactoros/blob/0eac0e66b09438b9a2a71d9071a466f0a55f3df2/src/RequestTrait.php#L55-L61

shadowhand: Has been that way since [1.0.0](https://github.com/zendframework/zend-diactoros/blob/1.0.0/src/RequestTrait.php#L68-L80).

weierophinney:
Thanks, @Nyholm

Nyholm:
Thank you for merging.

Issue #199, opened 2016-09-05
CallbackStream is incompatible with SapiStreamEmitter


markstory:
I'm not sure if this is intentional, but the current `CallbackStream` implementation is not compatible with `SapiStreamEmitter`. Because `CallbackStream` throws an exception on `rewind()` it cannot be combined with `SapiStreamEmitter`. A trivial example would be

```php
$fp = imagecreatefrompng('/path/to/meme.png');
$response = new Response();
$response = $response
  ->withHeader('Content-Type', 'image/png')
  ->withBody(new CallbackStream(function () use ($fp) {
    imagepng($fp);
    imagedestroy($fp);
  });

$emitter = new SapiStreamEmitter();
$emitter->emit($response);
```

Is there a reason that the `SapiStreamEmitter` is incompatible with `CallbackStream`, or should frameworks/applications be inspecting the response stream and choosing the correct emitter based stream type?

Related to cakephp/cakephp#9408

weierophinney:
> Is there a reason that the SapiStreamEmitter is incompatible with CallbackStream,

No, not really. :smile:

When I created the `CallbackStream`, I considered a `rewind()` operation as problematic; you can't "rewind" a code invocation. As such, I had it raise an exception.

There are two potential possibilities here:

- We update the `SapiStreamEmitter` to not call `rewind()` if it detects a `CallbackStream`.
- We update `CallbackStream` to no longer throw an exception on `rewind()`.

I'm okay with either, though probably more okay with the second; would you be willing to create a PR, @markstory ?

markstory:
Would be checking if the stream `isSeekable` before rewinding work for you? That's what I ended up doing in CakePHP as a temporary measure.

weierophinney:
Ooh! I like that even better! Go for it!

PR #186, opened 2016-07-05
Fixed spelling mistake lenght -> length

hannesvdvreken:

PR #200, opened 2016-09-07
Fix incompatibilities between Unseekable streams and SapiStreamEmitter

markstory:
Make `SapiStreamEmitter` work with streams that don't support seek operations. This allows `CallbackStream` responses to be combined with `SapiStreamEmitter`.

Refs #199

### DIFF
--- a/src/Response/SapiStreamEmitter.php
+++ b/src/Response/SapiStreamEmitter.php
@@ -76,10 +76,17 @@ class SapiStreamEmitter implements EmitterInterface
      */
     private function emitBodyRange(array $range, ResponseInterface $response, $maxBufferLength)
     {
-        list($unit, $first, $last, $lenght) = $range;
+        list($unit, $first, $last, $length) = $range;
 
         ++$last; //zero-based position
         $body = $response->getBody();
+
+        if (!$body->isSeekable()) {
+            $contents = $body->getContents();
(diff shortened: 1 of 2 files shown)


markstory:
Looks like your PHPCS checks are failing. Should I fix those up?

src/Response/SapiStreamEmitter.php
```diff
@@ -72,10 +76,17 @@ private function emitBody(ResponseInterface $response, $maxBufferLength)
      */
     private function emitBodyRange(array $range, ResponseInterface $response, $maxBufferLength)
     {
-        list($unit, $first, $last, $lenght) = $range;
+        list($unit, $first, $last, $length) = $range;
```
samsonasik: this change already at #186

samsonasik:
@markstory I created PR #201 for the cs fix

weierophinney:
Thanks for this, @markstory\!

markstory:
Thanks for the quick merge @weierophinney 💃

src/Response/SapiStreamEmitter.php
markstory: Oh, I didn't see that.

weierophinney: No worries; I merged in order, and since they were the exact same change, it presented no conflict during merge. :smile:

PR #205, opened 2016-09-07
Drop php 5.4 and 5.5 support

snapshotpl:


weierophinney:
No can do, at least not yet. We provide support for versions prior to 5.6 to accommodate the symfony ecosystem (as v2 introduced the PSR-7 bridge, and the last LTS still supports 5.4). We can look at this next year.

weierophinney:
I'm thinking what we'll do is one last minor version supporting pre-5.6 versions, and then mark that an LTS version that will only receive security and/or critical bugfixes. We can then do a 2.0 version sooner rather than later, and move any new features to that.

I'll come back to this in a few weeks so we can proceed.

test/CallbackStreamTest.php
```diff
@@ -89,13 +89,13 @@ public function testGetSize()
...
 
-    /**
-     * @expectedException RuntimeException
-     */
     public function testTell()
     {
         $stream = new CallbackStream(function () {
         });
+
+        $this->setExpectedException(RuntimeException::class);
```
samsonasik: use `expectException()` instead, as `setExpectedException()` is deprecated and will be removed in phpunit 6

snapshotpl:
@weierophinney a few weeks go away, what about this?

weierophinney:
Currently on vacation, @snapshotpl; will review next week.

test/Request/SerializerTest.php
```diff
@@ -266,36 +272,39 @@ public function messagesWithInvalidHeaders()
...
 
     public function testFromStreamThrowsExceptionWhenStreamIsNotReadable()
     {
-        $this->setExpectedException('InvalidArgumentException');
         $stream = $this
-            ->getMockBuilder('Psr\Http\Message\StreamInterface')
+            ->getMockBuilder(StreamInterface::class)
             ->getMock();
 
-        $stream->expects($this->once())->method('isReadable')
```
asgrim: Why remove the `expects`? Surely in these cases we want to ensure `isReadable` is only called once still? (same question applies to other similar changes). Only thing I can think of is that because an exception is thrown that the mock expectations are not checked? (I don't know off my head without checking)

test/Response/JsonResponseTest.php
```diff
@@ -70,22 +71,18 @@ public function testCanProvideAlternateContentTypeViaHeadersPassedToConstructor(
...
         $resource = fopen('php://memory', 'r');
+
+        $this->expectException(InvalidArgumentException::class);
+
         new JsonResponse($resource);
     }
 
     public function testJsonErrorHandlingOfBadEmbeddedData()
     {
-        if (version_compare(PHP_VERSION, '5.5', 'lt')) {
```
asgrim: Yuss! :+1:

test/CallbackStreamTest.php
```diff
@@ -89,13 +89,13 @@ public function testGetSize()
         $this->assertNull($ret);
     }
 
-    /**
-     * @expectedException RuntimeException
```
asgrim: Ideally we should try to avoid unrelated changes; as far as I know, there's nothing broken about `@expectedException` is there?

weierophinney:
I've changed my mind on this, and am scheduling this for 1.5.0.

composer.json
```diff
@@ -21,11 +21,11 @@
   },
   "require": {
     "php": "^5.6 || ^7.0",
-    "psr/http-message": "~1.0"
+    "psr/http-message": "^1.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "^4.6 || ^5.5",
-    "zendframework/zend-coding-standard": "~1.0.0",
+    "phpunit/phpunit": "^5.6 || ^6.0",
```
weierophinney: We need to get a bit more specific on these, as there's breakage in early versions of 6.0 and 5.6 that have often led to test failures in other repos. Try `^5.7.16 || ^6.0.8` here.

composer.json
```diff
@@ -21,11 +21,11 @@
...
+    "zendframework/zend-coding-standard": "^1.0",
```
weierophinney: This is one of the few that needs to use the `~` operator, as we don't want new PRs to suddenly fail due to new rules being added to the standard.

test/CallbackStreamTest.php
weierophinney: I'll allow it. Semantically, having the call to `expectException()` immediately prior to the code that raises the exception is better. While I realize it's unrelated, since this is updating the PHPUnit version already, we might as well do some cleanup.

test/MessageTraitTest.php
```diff
@@ -212,8 +216,9 @@ public function invalidHeaderValues()
      */
     public function testWithHeaderRaisesExceptionForInvalidValueType($value)
     {
-        $this->setExpectedException('InvalidArgumentException', 'Invalid header value');
-        $message = $this->message->withHeader('X-Foo', $value);
+        $this->expectException(InvalidArgumentException::class, 'Invalid header value');
```
weierophinney: This isn't doing what you think.

`expectException()` only accepts one argument, the exception type to test for. If you also want to test the message, you need to add an additional call to `expectExceptionMessage()`.

test/MessageTraitTest.php
```diff
@@ -229,8 +234,9 @@ public function testWithHeaderReplacesDifferentCapitalization()
      */
     public function testWithAddedHeaderRaisesExceptionForNonStringNonArrayValue($value)
     {
-        $this->setExpectedException('InvalidArgumentException', 'must be a string');
-        $message = $this->message->withAddedHeader('X-Foo', $value);
+        $this->expectException(InvalidArgumentException::class, 'must be a string');
```
weierophinney: Same comment here, and throughout; any two-argument calls to `setExpectedException()` need to be re-written as two discrete calls to `expectException()` and `expectExceptionMessage()`.

test/Request/SerializerTest.php
```diff
@@ -266,35 +272,28 @@ public function messagesWithInvalidHeaders()
...
-            ->getMock();
-
-        $stream->expects($this->once())->method('isReadable')
-            ->will($this->returnValue(true));
+        $stream = $this->createMock(StreamInterface::class);
+        $stream->method('isReadable')->will($this->returnValue(true));
+        $stream->method('isSeekable')->will($this->returnValue(false));
 
-        $stream->expects($this->once())->method('isSeekable')
-            ->will($this->returnValue(false));
```
weierophinney: I'm not sure I understand why these mocks were rewritten. The combination of `getMockBuilder()` + `getMock()` works in PHPUnit 5.6+ and 6.0+, as does the syntax of `expects()`. The changes you have introduced loosen the expectations by removing the `expects($this->once())` constraints...

test/Response/SerializerTest.php
```diff
@@ -194,35 +199,28 @@ public function messagesWithInvalidHeaders()
...
-        $this->setExpectedException('InvalidArgumentException');
-        $stream = $this
-            ->getMockBuilder('Psr\Http\Message\StreamInterface')
-            ->getMock();
-
-        $stream->expects($this->once())->method('isReadable')
-            ->will($this->returnValue(true));
+        $stream = $this->createMock(StreamInterface::class);
+        $stream->method('isReadable')->will($this->returnValue(true));
+        $stream->method('isSeekable')->will($this->returnValue(false));
```
weierophinney: Same comment here about the removal of `expects()`.

test/ServerIntegrationTest.php
```diff
@@ -43,5 +42,7 @@ public function testPassesBufferLevelToSapiStreamEmitter()
         );
         $server->setEmitter($emitter->reveal());
         $server->listen();
+
+        ob_end_clean();
```
weierophinney: Why was this added? Was it needed?

test/ServerRequestFactoryTest.php
```diff
@@ -20,6 +20,18 @@
...
+    protected function setUp()
+    {
+        $this->globalServer = $_SERVER;
+    }
+
+    protected function tearDown()
+    {
+        $_SERVER = $this->globalServer;
+    }
+
```
weierophinney: Why is this functionality needed? The `@backupGlobals` and `@preserveGlobalState` annotations are `enabled` by default, which should make this unnecessary. What did you observe that led to their addition?

test/ServerTest.php
```diff
@@ -268,17 +277,14 @@ public function testListenPassesCallableArgumentToCallback()
...
             ->method('getHeaders')
             ->will($this->returnValue([]));
 
         $responseBody = new Stream('php://temp');
         $this->response
-            ->expects($this->any())
             ->method('getBody')
             ->willReturn($responseBody);
         $this->response
-            ->expects($this->any())
```
weierophinney: The `expects()` calls need to be re-instated; these were there for a reason.

test/ServerIntegrationTest.php
weierophinney: Discovered why when removing locally; PHPUnit flags it as a risky test due to the fact it did not close its own output buffers.

test/ServerRequestFactoryTest.php
weierophinney: The test `testFromGlobalsUsesCookieSuperGlobalWhenCookieHeaderIsNotSet()` was causing inconsistent results due to superglobal state otherwise. This is better alleviated by the combination of `@runInSeparateProcess` and `@preserveGlobalState disabled` annotations.

weierophinney:
@snapshotpl I've pushed the changes I've requested to your branch at this time, and will rebase to incorporate them into your own commits once I can verify the changes pass continuous integration.

weierophinney:
Thanks, @snapshotpl

test/MessageTraitTest.php
snapshotpl: Good catch

PR #207, opened 2016-09-15
Remove fast_finish flag

Koopzington:
It has been confirmed that the fast_finish flag is responsible for triggering multiple build notifications on #zftalk.dev.


weierophinney:
Thanks, @Koopzington

PR #208, opened 2016-09-27
Add missing HTTP status codes/messages

ravage84:
At https://github.com/cakephp/cakephp/issues/9516, we were discussing missing status codes in CakePHP's and Zend's libraries.

This brings both (almost) [in sync](https://github.com/cakephp/cakephp/pull/9523) by adding missing status codes & messages.

Used https://httpstatuses.com/ as reference.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -44,6 +44,7 @@ class Response implements ResponseInterface
         206 => 'Partial Content',
         207 => 'Multi-status',
         208 => 'Already Reported',
+        226 => 'IM used',
         // REDIRECTION CODES
         300 => 'Multiple Choices',
         301 => 'Moved Permanently',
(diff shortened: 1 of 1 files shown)


src/Response.php
```diff
@@ -81,8 +83,10 @@ class Response implements ResponseInterface
         428 => 'Precondition Required',
```
inoas: 421 Misdirected Request is still missing.
Ref: https://httpstatuses.com/421
Ref: https://github.com/cakephp/cakephp/pull/9530/files#diff-a57bc7407c64664db661cb46c465f363R81

inoas:
Would you consider changing `306 => 'Switch Proxy', // Deprecated` to `306 => '(Unused)',`?
Ref: https://github.com/cakephp/cakephp/pull/9534

weierophinney:
Thanks, @ravage84. @ionas -- I incorporated your suggestions on merge.

PR #210, opened 2016-10-06
add composer.lock & update travis config

stefanotorresi:
as per zendframework/zendframework#7660

also entails these changes:

- make composer scripts entries consistent with the other components
- fix travis.yml env variables to not be affected by travis-ci/travis-ci#1444


weierophinney:
Thanks, @stefanotorresi. I cherry-picked to master.

PR #211, opened 2016-10-10
Add support for UTF-8 query string

danizord:
https://github.com/zendframework/zend-diactoros/pull/159 adds support for UTF-8 in host, and https://github.com/zendframework/zend-diactoros/pull/166 for path, but the query string still lacks UTF-8 support.


danizord:
So, I've fixed it by adding a `\u` suffix to the `filterQueryOrFragment()` regex

weierophinney:
Thanks, @danizord

PR #212, opened 2016-11-09
Actually accept string as body of ServerRequest

maxtsepkov:
The docblock and common sense suggest that ServerRequest constructor can accept $body as a string.
However doing so is not detected and code think that passed string is stream name.
I'm not sure if this is by design, it's a proposal, but in tests It's possible to pass a string as body.

Code suggested is just a quick scratch, elaboration needed.

Now I construct a `php://memory` stream and pass it to ServerRequest.

### DIFF
--- a/src/RequestTrait.php
+++ b/src/RequestTrait.php
@@ -64,7 +64,14 @@ trait RequestTrait
 
         $this->method = $method ?: '';
         $this->uri    = $this->createUri($uri);
-        $this->stream = $this->getStream($body, 'wb+');
+        try {
+            $this->stream = $this->getStream($body, 'wb+');
+        } catch (\InvalidArgumentException $e) {
+            $stream = fopen('php://memory', 'r+');
+            fwrite($stream, $body);
+            rewind($stream);
+            $this->stream = $stream;
+        }
 
         list($this->headerNames, $headers) = $this->filterHeaders($headers);
         $this->assertHeaders($headers);


src/RequestTrait.php
```diff
@@ -64,7 +64,14 @@ private function initialize($uri = null, $method = null, $body = 'php://memory',
 
         $this->method = $method ?: '';
         $this->uri    = $this->createUri($uri);
-        $this->stream = $this->getStream($body, 'wb+');
+        try {
+            $this->stream = $this->getStream($body, 'wb+');
+        } catch (\InvalidArgumentException $e) {
```
samsonasik: `InvalidArgumentException` already in use statement, no need `\` prefix

weierophinney:
This is not something we plan to support.

The constructor allows a string _only for purposes of providing a stream resource_. In other words, the string provided should be a file name or a valid stream descriptor that can be passed to `fopen()`.

In the future, we plan to support PSR-17, which will provide factories for creating any PSR-7 type. PSR-17 is actually providing several factories around streams:

- one for creating an instance from an actual resource.
- one for creating an instance from a resource identifier (file name, stream descriptor, etc.).
- one for creating an instance from a string, which will implicitly create a `php://temp` resource populated with those contents.

While we do not plan to change the constructor in our own representation, we *could* have a static method, `Stream::fromString()` for that purpose. If you would like to alter this patch to do that instead, I'll review it again when those changes are done.

weierophinney:
Closing. If somebody would like to take the suggestion I made in my previous comment and create a new pull request, we'll review it.

PR #213, opened 2016-11-13
New coding standard

Koopzington:
Usage of the newly introduced zendframework/zend-coding-standard for CS checks.


weierophinney:
Thanks, @Koopzington

PR #215, opened 2016-11-25
Added a getter for the temp file-name

Supersander:
I want to be able to access the temp file-name of an uploaded file.

The specific case where i ran into this was handling a uploaded file to be imported by PHPExcel. PHPExcel only allows loading files from a filename and not from a stream for example.
I have no interest in writing the file away when i dont have to and no interest in keeping the file after the request has been handled.
Though you could argue the is a PHPExcel problem (and I agree), I see no reason to not allow a getter for the temp-file-name.

### DIFF
--- a/src/UploadedFile.php
+++ b/src/UploadedFile.php
@@ -219,6 +219,14 @@ class UploadedFile implements UploadedFileInterface
     public function getClientMediaType()
     {
         return $this->clientMediaType;
+    }   
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getFile()
+    {
+        return $this->file;
     }
 
     /**


Ocramius:
Seems wrong to me. If the limitation is in a different API, dumping to another file and then passing the path to that file further on is perfectly legit. It's inefficient, but keeps th API cleaner, especially considering that you may not have temporary paths with some http servers, whereas allowing streams in PHPExcel would solve the problem completely.

/cc @MarkBaker @PowerKiki

PowerKiKi:
I still don't know well enough the code base to know for sure if it would be technically possible, and desirable, to support stream. But one thing is for sure, if one of you would like to contribute to the project, then please do so in https://github.com/PHPOffice/PhpSpreadsheet, and not PHPExcel, ([see why](https://github.com/PHPOffice/PhpSpreadsheet#phpexcel-vs-phpspreadsheet-)).

Ocramius:
Meanwhile, there is a separate approach to deal with this, although ugly and complicated: use a stream wrapper that exposes a file path, then use that path in PHPExcel.

I am not merging additional hacky public API here though, so I'm closing the PR. Feel free to continue discussing here, if needed.

PR #219, opened 2016-12-28
Serialize into array

snapshotpl:
For me it's useful when I want to serialize messages to log using psr-4 (into context)


snapshotpl:
If it ok I will make serializer for response. What are you think about deserializers from arrays?

Ocramius:
@snapshotpl would segregate it to a separate serializer or interface (implementation can be added to existing one, but ISP must be respected IMO)

snapshotpl:
@Ocramius as it's static method interface doesn't sense here imo.

Ocramius:
Misread it, good point. Indeed, statics don't need interfaces.

snapshotpl:
What about array deserializer idea?

weierophinney:
I would create `Zend\Diactoros\Request\ArraySerializer` and `Zend\Diactoros\Response\ArraySerializer` classes, each with the methods:

- `fromArray(array $messageData)`
- `toArray(RequestInterface $request)`

The purpose of the current serializers is specifically for string de/serialization, and the methods exposed reflect that purpose. New classes specifically for array de/serialization make sense here.

snapshotpl:
@weierophinney @Ocramius ready to review

weierophinney:
Merged to develop for release with 1.4.0; thanks, @snapshotpl

Issue #221, opened 2016-12-31
SapiStreamEmitter shouldn't emitBodyRange for non-byte units


doublecompile:
The `SapiStreamEmitter` calls the `emitBodyRange` method any time the response's `Content-Range` header can be parsed into *unit* *first*-*last*/*length*. However, this method should only be called if the *unit* is `bytes`. 

Check out [this StackOverflow question](http://stackoverflow.com/questions/1434647/using-the-http-range-header-with-a-range-specifier-other-than-bytes). RFC 2616 allows for non-byte units. In fact, as noted in the question, Dojo Toolkit and Dstore specifically use the `items` unit to note subsets of JSON REST results. (e.g. an application can return `Content-Range: items 0-0/1` for a single record).

If I send a JSON array and include this header, the `SapiStreamEmitter` will truncate the response to:

```
[
```

I can prepare a pull request in a few moments.

Ocramius:
@doublecompile is this supposed to be closed?

doublecompile:
@Ocramius Yup! It was merged in #222 and @weierophinney released it in 1.3.8.

PR #222, opened 2016-12-31
SapiStreamEmitter only sends subsets if the unit is bytes

doublecompile:
Resolves issue #221


weierophinney:
Thanks, @doublecompile\!

Issue #225, opened 2017-01-21
JsonResponse not working


Wikiki:
Hello,

I try to use the JsonResponse but without success.
I use the following code to send an array as response content
`return new JsonResponse( $data, $status );`

Then I use the following code to emit the response
```
$emitter = new SapiStreamEmitter();
$emitter->emit( $response );
```

But in my browser the response is empty. I always get a HTTP Status code 200 with an empty body. Even if I force status code to 201 (for example).

Can you help me ?

Regards.

fcabralpacheco:
I wasn't able to reproduce the issue in the tests.

The behavior appears to be headers already sent and display_errors = Off.

What is your error reporting configuration (in phpinfo(), error reporting and display_errors directives)?

Wikiki:
Hi,

the directives are the following:
display_errors = Off
error_reporting = 32767

Wikiki:
Hi again,

You're right...if I put display_errors to on then I got a PHP error. After fixing it it's working perfectly.
Thanks for your help.

PR #226, opened 2017-01-22
Improves SapiStreamEmitter tests, preventing unnecessary string casts

fcabralpacheco:
@Ocramius 

I tried to improve some points in the SapiStreamEmitter tests:

- Reduced code repetition by creating a new method for the Stream Prophecy common code.

- Added a method for expanding the test data set. This facilitates the addition of new data set with one information changed (e.g.: the contents).

- Improved checks of methods that can or cannot be called.

- Removed  the expectation of read counts because the buffer size is the maximum, the implementation can use a smaller one if you want.

- Considering bug #225, which I was not able to reproduce, both in the test and in an Expressive application. Added new tests for each type of response, checking what was emitted, and whether the content type was not changed.

Just a few comments:

- I didn't add out of range conditions to the tests (in testEmitRangeStreamResponse). I think that the application should handle this type of case, and it may be not possible to detect this type of condition in non seekable bodies.

- Please keep prophecy prediction after reveal. The Prophecy don't calls promise if it signature don't match exact with prediction signature.
E.g.:  $stream->seek(Argument::type('integer'), Argument::any()); vs $stream->seek($first)->should($seekPredictionClosure); and $ stream->seek($first, SEEK_SET)->should($seekPredictionClosure);
This occurs because signatures generate three different prophecies, one with a promise and two with a prediction, on execution only highest rated prophecy is executed. If you pass the "$first" value, only first prediction would be executed.

Linked to #224, #223

### DIFF
--- a/test/Response/SapiStreamEmitterTest.php
+++ b/test/Response/SapiStreamEmitterTest.php
@@ -14,6 +14,7 @@ use Zend\Diactoros\CallbackStream;
 use Zend\Diactoros\Response;
 use Zend\Diactoros\Response\SapiStreamEmitter;
 use ZendTest\Diactoros\TestAsset\HeaderStack;
+use Zend\Diactoros\Response\JsonResponse;
 
 class SapiStreamEmitterTest extends SapiEmitterTest
 {
@@ -57,52 +58,146 @@ class SapiStreamEmitterTest extends SapiEmitterTest
         }
     }
 
-    public function emitBodyProvider()
+    private function expandTestData()
     {
-        return [
-            [true,   '01234567890123456789'    , 10, 2],
-            [true,   '012345678901234567890123', 10, 3],
-            [false,  '01234567890123456789'    , 10, 2],
-            [false,  '012345678901234567890123', 10, 3],
-            [true,   '012345678901234567890123', 0, 0],
-            [false,  '012345678901234567890123', 0, 0],
-        ];
-    }
+        $initialData = func_get_args();
 
-    /**
-     * @dataProvider emitBodyProvider
-    */
-    public function testEmitBody($seekable, $contents, $maxBufferLength, $expectedReads = 0)
-    {
-        $position = 0;
+        $initialData = array_filter($initialData, 'is_array');
 
-        $closureContents = function () use (& $contents, & $position) {
-            $position = strlen($contents);
-            return $contents;
+        $resultData = [];
+
+        $closureExpand = function ($item, $key, $parameters) use (& $closureExpand) {
+            if (! is_array($item)) {
+                $item = [$item];
+            }
+
+            $parameters['items'] = array_merge($parameters['items'], $item);
+
+            if (count($parameters['initialData']) > 0) {
+                array_walk(array_shift($parameters['initialData']), $closureExpand, $parameters);
+            } else {
+                $parameters['resultData'][] = $parameters['items'];
+            }
         };
 
+        array_walk(array_shift($initialData), $closureExpand, [
+            'initialData' => $initialData,
+            'items' => [],
+            'resultData' => & $resultData]);
+
+        return $resultData;
+    }
+
+    private function getStreamProphecy(& $contents, & $size, & $position, & $peakBufferLength)
+    {
         $stream = $this->prophesize('Psr\Http\Message\StreamInterface');
-        $stream->getSize()->willReturn(strlen($contents));
-        $stream->isSeekable()->willReturn($seekable);
-        $stream->isReadable()->willReturn($expectedReads > 0 ? true : false);
-        $stream->__toString()->will($closureContents);
-        $stream->getContents()->will($closureContents);
+
+        $stream->__toString()->will(function () use (& $contents, & $position, $size) {
+            if (is_callable($contents)) {
+                $data = $contents(0);
+            } else {
+                $data = $contents;
+            }
+
+            $position = $size;
+
+            return $data;
+        });
+
+        $stream->getSize()->willReturn($size);
+
+        $stream->tell()->will(function () use (& $position) {
+            return $position;
+        });
+
+        $stream->eof()->will(function () use ($size, & $position) {
+            return ($position >= $size);
+        });
+
+        $stream->seek(Argument::type('integer'), Argument::any())->will(function ($args) use ($size, & $position) {
+            if ($args[0] < $size) {
+                $position = $args[0];
+                return true;
+            }
+
+            return false;
+        });
 
         $stream->rewind()->will(function () use (& $position) {
             $position = 0;
             return true;
         });
 
-        $stream->eof()->will(function () use (& $contents, & $position) {
-            return ! isset($contents[$position]);
+        $stream->read(Argument::type('integer'))
+            ->will(function ($args) use (& $contents, & $position, & $peakBufferLength) {
+                if (is_callable($contents)) {
+                    $data = $contents($position, $args[0]);
+                } else {
+                    $data = substr($contents, $position, $args[0]);
+                }
+
+                if ($args[0] > $peakBufferLength) {
+                    $peakBufferLength = $args[0];
+                }
+
+                $position += strlen($data);
+
+                return $data;
+            });
+
+        $stream->getContents()->will(function () use (& $contents, & $position) {
+            if (is_callable($contents)) {
+                $remainingContents = $contents($position);
+            } else {
+                $remainingContents = substr($contents, $position);
+            }
+
+            $position += strlen($remainingContents);
+
+            return $remainingContents;
         });
 
-        $stream->read(Argument::type('integer'))->will(function ($args) use (& $contents, & $position) {
-            $data      = substr($contents, $position, $args[0]);
-            $position += strlen($data);
+        return $stream;
+    }
 
-            return $data;
-        });
+    public function emitStreamResponseProvider()
+    {
+        $capabilities = [
+            [true, true],
+            [true, false],
+            [false, true],
+            [false, false],
+        ];
+
+
+        $contents = [
+            ['01234567890987654321'],
+            ['01234567890987654321012'],
+        ];
+
+        $bufferLengths = [
+          [10],
+          [20],
+          [100],
+        ];
+
+        return $this->expandTestData($capabilities, $contents, $bufferLengths);
+    }
+
+    /**
+     * @dataProvider emitStreamResponseProvider
+     */
+    public function testEmitStreamResponse($seekable, $readable, $contents, $maxBufferLength)
+    {
+        $size = strlen($contents);
+        $position = 0;
+        $peakBufferLength = 0;
+        $rewindCalled = false;
+        $fullContentsCalled = false;
+
+        $stream = $this->getStreamProphecy($contents, $size, $position, $peakBufferLength);
+        $stream->isSeekable()->willReturn($seekable);
+        $stream->isReadable()->willReturn($readable);
 
         $response = (new Response())
             ->withStatus(200)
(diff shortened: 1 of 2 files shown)


test/Response/SapiStreamEmitterTest.php
```diff
@@ -57,76 +61,322 @@ public function testDoesNotInjectContentLengthHeaderIfStreamSizeIsUnknown()
...
+            'resultData' => & $resultData]);
+
+        return $resultData;
+    }
+
+    private function getStreamProphecy(& $contents, & $size, & $position, & $peakBufferLength)
+    {
+        $stream = $this->prophesize('Psr\Http\Message\StreamInterface');
+
+        $stream->__toString()->will(function () use (& $contents, & $position, $size) {
```
Ocramius: Is `$size` supposed to *not* be passed in by reference?

test/Response/SapiStreamEmitterTest.php
```diff
@@ -57,76 +61,322 @@ public function testDoesNotInjectContentLengthHeaderIfStreamSizeIsUnknown()
...
+            $parameters['items'] = array_merge($parameters['items'], $item);
+
+            if (count($parameters['initialData']) > 0) {
+                array_walk(array_shift($parameters['initialData']), $closureExpand, $parameters);
+            } else {
+                $parameters['resultData'][] = $parameters['items'];
+            }
+        };
+
+        array_walk(array_shift($initialData), $closureExpand, [
```
Ocramius: Please use consistent alignment (PSR-2)

test/Response/SapiStreamEmitterTest.php
```diff
@@ -57,76 +61,322 @@ public function testDoesNotInjectContentLengthHeaderIfStreamSizeIsUnknown()
...
-            [true,   '012345678901234567890123', 10, 3],
-            [false,  '01234567890123456789'    , 10, 2],
-            [false,  '012345678901234567890123', 10, 3],
+        $initialData = func_get_args();
+
+        $initialData = array_filter($initialData, 'is_array');
+
+        $resultData = [];
+
+        $closureExpand = function ($item, $key, $parameters) use (& $closureExpand) {
```
Ocramius: What does `closureExpand` mean here?

test/Response/SapiStreamEmitterTest.php
```diff
@@ -57,76 +61,322 @@ public function testDoesNotInjectContentLengthHeaderIfStreamSizeIsUnknown()
...
+
+        array_walk(array_shift($initialData), $closureExpand, [
+            'initialData' => $initialData,
+            'items' => [],
+            'resultData' => & $resultData]);
+
+        return $resultData;
+    }
+
+    private function getStreamProphecy(& $contents, & $size, & $position, & $peakBufferLength)
```
Ocramius: s/get/build

Ocramius: The act that all this stuff is by-ref makes it extremely hard to read

test/Response/SapiStreamEmitterTest.php
```diff
@@ -57,76 +61,322 @@ public function testDoesNotInjectContentLengthHeaderIfStreamSizeIsUnknown()
...
+
+            $position += strlen($remainingContents);
+
+            return $remainingContents;
+        });
+
+        return $stream;
+    }
+
+    public function emitStreamResponseProvider()
```
Ocramius: Please document the return type o this data provider

test/Response/SapiStreamEmitterTest.php
```diff
@@ -57,76 +61,322 @@ public function testDoesNotInjectContentLengthHeaderIfStreamSizeIsUnknown()
...
+            [false,  true,  1000,   20,       null],
+            [false, false,   100,  320,       null],
+            [true,   true,  1000,   20,   [25, 75]],
+            [false,  true,  1000,   20,   [25, 75]],
+            [true,   true,  1000,   20, [250, 750]],
+            [false,  true,  1000,   20, [250, 750]],
+        ];
+
+        $bufferLengths = [
+                [512],
```
Ocramius: Alignment broken here

test/Response/SapiStreamEmitterTest.php
```diff
@@ -57,76 +61,322 @@ public function testDoesNotInjectContentLengthHeaderIfStreamSizeIsUnknown()
...
+        $this->assertEquals($seekable, $rewindCalled);
+        $this->assertEquals(! $readable, $fullContentsCalled);
+        $this->assertEquals($contents, $emittedContents);
+        $this->assertLessThanOrEqual($maxBufferLength, $peakBufferLength);
+    }
+
+    public function emitRangeStreamResponseProvider()
+    {
+        $capabilities = [
+                [true,   true],
```
Ocramius: Alignment broken here

test/Response/SapiStreamEmitterTest.php
```diff
@@ -211,70 +445,80 @@ function () use (& $closureTrackMemoryUsage) {
...
+        $this->assertEquals($contents, ob_get_clean());
+    }
 
-        $stream->seek(Argument::type('integer'), Argument::any())->will(function ($args) use (& $position) {
-            $position = $args[0];
-            return true;
-        });
+    public function emitJsonResponseProvider()
+    {
+        return [[0.1                                                                         ],
```
Ocramius: Alignment broken here

test/Response/SapiStreamEmitterTest.php
```diff
@@ -211,70 +445,80 @@ function () use (& $closureTrackMemoryUsage) {
...
+        $this->assertEquals($contents, ob_get_clean());
     }
 
     public function contentRangeProvider()
     {
         return [
-            ['bytes 0-2/*', 'Hello world', 'Hel'],
-            ['bytes 3-6/*', 'Hello world', 'lo w'],
-            ['items 0-0/1', 'Hello world', 'Hello world'],
+                ['bytes 0-2/*', 'Hello world', 'Hel'],
```
Ocramius: Alignment broken here

test/Response/SapiStreamEmitterTest.php
```diff
@@ -57,76 +61,322 @@ public function testDoesNotInjectContentLengthHeaderIfStreamSizeIsUnknown()
         }
     }
 
-    public function emitBodyProvider()
+    private function expandTestData()
```
Ocramius: While I get that you wanted to normalize the mocking done for the data providers, this method has become absolutely unreadable. I would suggest duplicating or the supported scenarios. Specifically based on:

 * number of passed in parameters
 * types of passed in parameters

test/Response/SapiStreamEmitterTest.php
```diff
@@ -57,76 +61,322 @@ public function testDoesNotInjectContentLengthHeaderIfStreamSizeIsUnknown()
...
+            $stream->seek(0, SEEK_SET)->should($rewindPredictionClosure);
+        } else {
+            $stream->rewind()->shouldNotBeCalled();
+            $stream->seek(Argument::type('integer'), Argument::any())->shouldNotBeCalled();
+        }
 
-            return $data;
-        });
+        if ($readable) {
+            $stream->__toString()->shouldNotBeCalled();
```
Ocramius: Most probably the important assertion that we missed before :S

test/Response/SapiStreamEmitterTest.php
```diff
@@ -57,76 +61,322 @@ public function testDoesNotInjectContentLengthHeaderIfStreamSizeIsUnknown()
...
+        ];
+
+        return $this->expandTestData($capabilities, $ranges, $contents, $bufferLengths);
+    }
+
+
+    /**
+     * @dataProvider emitRangeStreamResponseProvider
+     */
+    public function testEmitRangeStreamResponse($seekable, $readable, array $range, $contents, $maxBufferLength)
```
Ocramius: Please document the parameter types/formats

test/Response/SapiStreamEmitterTest.php
fcabralpacheco: I get it. I'll use duplication.
I thought the same when I finished writing this method.
I tried a version without array_walk (using for, next, current), it got worse.

test/Response/SapiStreamEmitterTest.php
fcabralpacheco: The method expands data like a tree. In the end of branch resides one data set.

test/Response/SapiStreamEmitterTest.php
fcabralpacheco: Ok. The $size don't need by ref, the $contents too (I thought it would be necessary at first due to memory usage).

The $position and $peakBufferLength is necessary, I can return them in an array. Like  this:
private function getStreamProphecy($contents, $size)
...
return [
   $stream,
   & $position,
   & $peakBufferLength,
];


And I can rename getStreamProphecy to setUpStreamProphecy.

What do you think?

test/Response/SapiStreamEmitterTest.php
fcabralpacheco: The $size don't need by ref.

test/Response/SapiStreamEmitterTest.php
fcabralpacheco: Okay. It was supposed to be just a reordering of this piece of the code. I messed things.

test/Response/SapiStreamEmitterTest.php
fcabralpacheco: @Ocramius  
I'm not sure if after the changes this still necessary.
Also, I don't know exactly how to document elements of an array according to the ZF CS (as far as I know. There is no official way to do this in phpdoc).

test/Response/SapiStreamEmitterTest.php
fcabralpacheco: I agree with you.

test/Response/SapiStreamEmitterTest.php
fcabralpacheco: Removed unecessary by-ref and replaced $peakBufferLength by a callable.
I think it's now become more intuitive.

test/Response/SapiStreamEmitterTest.php
Ocramius: Oh, it just needed a human-friendly description. The new syntax makes it much more approachable, so it's not needed anymore.

Ocramius:
Backported into `master` via zendframework/zend-diactoros@c2c59823f86b9d457794bb829823f4b3271dabdb

fcabralpacheco:
@Ocramius 

I'm sorry, just a little type correction/type hinting.

https://github.com/fcabralpacheco/zend-diactoros/commit/346e468b1b65e8e085fb19a4a409b3db803616ac

Thanks.

Ocramius:
@fcabralpacheco cherry-picking it, thanks!

Ocramius:
@fcabralpacheco also, sorry for having to do all this because of a messed up CS fix by me: thanks for all the patience!

fcabralpacheco:
I'm sorry again, just a little correction.

https://github.com/fcabralpacheco/zend-diactoros/commit/555e9cd0e0eedb886c34bc37d412b1b5020d0c2e

https://github.com/fcabralpacheco/zend-diactoros/commit/e4067691d6db8dd51a81960ddbad7137ef4f67de

Thanks

Ocramius:
Already applied https://github.com/fcabralpacheco/zend-diactoros/commit/555e9cd0e0eedb886c34bc37d412b1b5020d0c2e

fcabralpacheco:
And thanks for all the patience too.

PR #230, opened 2017-02-07
Fixes typo IM Used response reason phrase typo

samsonasik:
uses uppercase for "used", changed to "Used"

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -44,7 +44,7 @@ class Response implements ResponseInterface
         206 => 'Partial Content',
         207 => 'Multi-status',
         208 => 'Already Reported',
-        226 => 'IM used',
+        226 => 'IM Used',
         // REDIRECTION CODES
         300 => 'Multiple Choices',
         301 => 'Moved Permanently',


Ocramius:
Ref: https://github.com/zendframework/zend-diactoros/pull/208

Ocramius:
@samsonasik is there any way to test these status codes from an RFC list or such?

samsonasik:
@Ocramius not sure..., probably use ReflectionProperty ?

Ocramius:
@samsonasik yeah, that part is indeed trivial, but we need some sort of authoritative list to compare against, or else we'll keep having these failures, especially if the spec is extended

sharifzadesina:
@Ocramius https://tools.ietf.org/html/rfc3229#section-10.4.1

Ocramius:
Yes, and is there no parseable version somewhere? Going paragraph per
paragraph seems silly

sharifzadesina:
@Ocramius I have linked the section with the URL, section 10.4.1

Ocramius:
Yes, and we'd need that for every paragraph 😅

sharifzadesina:
@Ocramius 
https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
https://github.com/elegantweb/framework/blob/master/src/Elegant/Http/Message/Response.php#L35-L94

Ocramius:
Must be an authoritative source of truth, not another implementation.

Anyway, I'm nitpicking: if this isn't feasible or practical, then this is
ready as-is

sharifzadesina:
@Ocramius Have you ever seen the WikiPedia page? implementation is just for making your job easier ;)
anyway you need also change status code 505 and 207.

Ocramius:
Eh, Wikipedia is not the IETF, although it's probably stable enough

sharifzadesina:
@Ocramius Yes, but you can find RFC numbers for each standard status code.

fcabralpacheco:
@Ocramius Accurate, updated (e.g. 408, 413, 414, 416, 505) and ease to parse (xml).

http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
http://www.iana.org/assignments/http-status-codes/http-status-codes.xml
http://www.iana.org/assignments/http-status-codes/http-status-codes.txt

However, is not perfect, some codes are missing (maybe unofficial or Easter egg - 418)

weierophinney:
Thanks, @samsonasik

PR #234, opened 2017-02-19
Adds response reason phrases test against IANA.

fcabralpacheco:
@Ocramius 

This is a possible implementation of response reason phrases test against the registry maintained by IANA.

I chose XML because of the possibility of validating before starting the test.

The test is marked as incomplete in case of invalid XML.

I'm not sure how is the best way to update it, but I believe a pre-release script would be the best alternative.

Something like this:

curl -o test/TestAsset/http-status-codes.xml http://www.iana.org/assignments/http-status-codes/http-status-codes.xml

About getting the xml when the test runs, I believe is overkill (and adds dependency of external resource), and reminds me of cases of service misuse.

One last comment, the IANA registry is the authoritative list for HTTP status codes (RFC7231, Section 8.2.)

Ref: #230, #208


test/TestAsset/http-status-codes.xml
```diff
@@ -0,0 +1,368 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
Ocramius: Where does this file come from? And can we instead request it via HTTPS? That's what I was asking for :-)

Ocramius: Specifically, the test should just download them by pinging https://www.iana.org/assignments/http-status-codes/http-status-codes.xml

fcabralpacheco: @Ocramius 

Changing the file name by URL in ResponseTest.php:77,  works just fine.

However, this can be a problem.

The test will depend on an external online resource.

And besides this, imagine if all libraries decide to do the same thing, requesting the file when the test runs.

I believe this may be end up leading to a case of service misuse.

One solution would be to add a pre-release script, something like this in composer.json ("scripts" block).

   "pre-release-tasks": [
      "@update-test-asset"
    ],
    "update-test-asset": "curl -o test/TestAsset/http-status-codes.xml https://www.iana.org/assignments/http-status-codes/http-status-codes.xml"

What do you think?

Ocramius: > I believe this may be end up leading to a case of service misuse.

One http request a day will not trigger misuse :-)

 > One solution would be to add a pre-release script, something like this in composer.json ("scripts" block).
 > 
 > "pre-release-tasks": [
 > "@update-test-asset"
 > ],
 > "update-test-asset": "curl -o test/TestAsset/http-status-codes.xml https://www.iana.org/assignments/http-status-codes/http-status-codes.xml"
 > 
 > What do you think?

If it's not automated, it's not checked. Let's please push it to the test suite, and run it every time.

fcabralpacheco: OK. I understand.

Check new version, I may remove "last modified" stuff to make it more simpler.

Apparently IANA server don't support "If-Modified-Since" (or I made some mistakes).

But. I think what timeout and fallback to file is necessary.

samsonasik:
@fcabralpacheco could you create alternate PR against zend-http also in favor of https://github.com/zendframework/zend-http/pull/110 ?

test/TestAsset/http-status-codes.xml
fcabralpacheco: @Ocramius 

I removed unused code. And I tried to make it simpler and clear.

fcabralpacheco:
@samsonasik Yes. I will create an alternate PR.

test/ResponseTest.php
```diff
@@ -66,29 +66,23 @@ public function testReasonPhraseDefaultsToStandards()
...
 
-    protected function checkNewVersionIanaHttpStatusCodes(\DOMDocument $ianaHttpStatusCodes)
+    protected function updateAndLoadIanaHttpStatusCodes()
     {
         set_error_handler(function ($errno, $errstr) {
             throw new \ErrorException($errstr, 0, $errno);
         });
 
-        $updateError = null;
+        $ianaHttpStatusCodes = new \DOMDocument();
```
Ocramius: requires `ext-xml` in the dev dependencies

test/ResponseTest.php
```diff
@@ -97,69 +91,59 @@ protected function checkNewVersionIanaHttpStatusCodes(\DOMDocument $ianaHttpStat
...
                 if ($validXml) {
                     file_put_contents(__DIR__ . '/TestAsset/http-status-codes.xml', $contents);
                     print 'IANA "http-status-codes.xml" updated successful' . "\n";
-                } else {
-                    $ianaHttpStatusCodes->load(__DIR__ . '/TestAsset/http-status-codes.xml');
                 }
             }
         } catch (\Exception $e) {
-            $updateError = $e->getMessage();
+            $errorMessage = $e->getMessage();
```
Ocramius: Remove this logic (the entire try/catch)

test/TestAsset/http-status-codes.xml
Ocramius: This file is to be removed from the test suite

test/TestAsset/iana-registry.rng
```diff
@@ -0,0 +1,198 @@
+<?xml version='1.0'?>
```
Ocramius: file to be removed from the PR

test/ResponseTest.php
```diff
@@ -66,6 +66,122 @@ public function testReasonPhraseDefaultsToStandards()
...
+        set_error_handler(function ($errno, $errstr) {
+            throw new \ErrorException($errstr, 0, $errno);
+        });
+
+        $ianaHttpStatusCodes = new \DOMDocument();
+        $validXml = false;
+        $errorMessage = null;
+        $httpStatus = 0;
+
+        try {
```
Ocramius: to be removed

test/ResponseTest.php
```diff
@@ -66,6 +66,122 @@ public function testReasonPhraseDefaultsToStandards()
...
+            $options = [
```
Ocramius: Doesn't need to be in a variable

test/ResponseTest.php
```diff
@@ -66,6 +66,122 @@ public function testReasonPhraseDefaultsToStandards()
...
+                ],
+            ];
+
+            $contents = file_get_contents(
+                'https://www.iana.org/assignments/http-status-codes/http-status-codes.xml',
+                false,
+                stream_context_create($options)
+            );
+
+            if ($http_response_header) {
```
Ocramius: This variable wasn't defined before

test/ResponseTest.php
```diff
@@ -66,6 +66,122 @@ public function testReasonPhraseDefaultsToStandards()
         $this->assertEquals('Unprocessable Entity', $response->getReasonPhrase());
     }
 
+    protected function updateAndLoadIanaHttpStatusCodes()
+    {
+        set_error_handler(function ($errno, $errstr) {
```
Ocramius: Remove this block - phpunit has a flag for raising exceptions on notices

test/ResponseTest.php
```diff
@@ -66,6 +66,122 @@ public function testReasonPhraseDefaultsToStandards()
...
+                stream_context_create($options)
+            );
+
+            if ($http_response_header) {
+                if (preg_match('/^HTTP\/[0-9\.]+\s*([0-9]+)\s*.+$/i', $http_response_header[0], $matches) > 0) {
+                    $httpStatus = $matches[1];
+                }
+            }
+
+            if ($httpStatus == 200) {
```
Ocramius: `if (200 !== $httpStatus) { self::fail(...); }`

test/TestAsset/iana-registry.rng
Ocramius: Hmm, never mind, this seems to be required for validity checks

test/ResponseTest.php
```diff
@@ -66,6 +66,122 @@ public function testReasonPhraseDefaultsToStandards()
...
+
+            try {
+                $ianaHttpStatusCodes->load(__DIR__ . '/TestAsset/http-status-codes.xml');
+                $validXml = $ianaHttpStatusCodes->relaxNGValidate(__DIR__ . '/TestAsset/http-status-codes.rng');
+            } catch (\Exception $e) {
+                $errorMessage = $e->getMessage();
+            }
+        }
+
+        restore_error_handler();
```
Ocramius: remove

test/ResponseTest.php
```diff
@@ -66,6 +66,122 @@ public function testReasonPhraseDefaultsToStandards()
...
+
+        if (! $validXml) {
+            if ($errorMessage) {
+                print 'Error on IANA "http-status-codes.xml" update. Error: ' . $errorMessage . "\n";
+            }
+
+            try {
+                $ianaHttpStatusCodes->load(__DIR__ . '/TestAsset/http-status-codes.xml');
+                $validXml = $ianaHttpStatusCodes->relaxNGValidate(__DIR__ . '/TestAsset/http-status-codes.rng');
+            } catch (\Exception $e) {
```
Ocramius: remove - let it bubble up

test/ResponseTest.php
```diff
@@ -66,6 +66,122 @@ public function testReasonPhraseDefaultsToStandards()
...
+                }
+            }
+        } catch (\Exception $e) {
+            $errorMessage = $e->getMessage();
+        }
+
+
+        if (! $validXml) {
+            if ($errorMessage) {
+                print 'Error on IANA "http-status-codes.xml" update. Error: ' . $errorMessage . "\n";
```
Ocramius: `self::fail()` is sufficient :-\

test/ResponseTest.php
```diff
@@ -66,6 +66,122 @@ public function testReasonPhraseDefaultsToStandards()
...
+                $validXml = $ianaHttpStatusCodes->relaxNGValidate(__DIR__ . '/TestAsset/http-status-codes.rng');
+            } catch (\Exception $e) {
+                $errorMessage = $e->getMessage();
+            }
+        }
+
+        restore_error_handler();
+
+        if (! $validXml) {
+            $this->markTestIncomplete(
```
Ocramius: `self::fail()` is better here.

test/ResponseTest.php
```diff
@@ -66,6 +66,122 @@ public function testReasonPhraseDefaultsToStandards()
...
+        }
+
+        return $ianaHttpStatusCodes;
+    }
+
+    public function ianaCodesReasonPhrasesProvider()
+    {
+        $ianaHttpStatusCodes = $this->updateAndLoadIanaHttpStatusCodes();
+
+        if (! $ianaHttpStatusCodes) {
```
Ocramius: The method should never return `null`

test/ResponseTest.php
```diff
@@ -66,6 +66,122 @@ public function testReasonPhraseDefaultsToStandards()
...
+        $xpath = new \DomXPath($ianaHttpStatusCodes);
+        $xpath->registerNamespace('ns', 'http://www.iana.org/assignments');
+
+        $records = $xpath->query('//ns:record');
+
+        foreach ($records as $record) {
+            $value = $xpath->query('.//ns:value', $record)->item(0)->nodeValue;
+            $description = $xpath->query('.//ns:description', $record)->item(0)->nodeValue;
+
+            if ($description === 'Unassigned' || $description === '(Unused)') {
```
Ocramius: `in_array()`

test/ResponseTest.php
```diff
@@ -66,6 +66,122 @@ public function testReasonPhraseDefaultsToStandards()
...
+                continue;
+            }
+
+            $range = preg_match('/^([0-9]+)\s*\-\s*([0-9]+)$/', $value, $matches);
+
+            if (! $range) {
```
Ocramius: Swap the conditional to avoid the negation. Also, `$range` variable is redundant

test/ResponseTest.php
froschdesign: @Ocramius 
http://php.net/manual/en/reserved.variables.httpresponseheader.php

test/ResponseTest.php
fcabralpacheco: Requires ext-dom.

fcabralpacheco:
@Ocramius

Removed file fallback  and error handling logic.

PHPUnit handles all errors/warnings/notices nicely, just self:fail() worked very well.

fcabralpacheco:
@Ocramius 

I have not updated the composer.lock file. Do I need to do this?

test/ResponseTest.php
```diff
@@ -66,6 +66,63 @@ public function testReasonPhraseDefaultsToStandards()
...
+                        'timeout' => 30,
+                    ],
+                ]
+            )
+        );
+
+        $ianaHttpStatusCodes->load('https://www.iana.org/assignments/http-status-codes/http-status-codes.xml');
+
+        if (! $ianaHttpStatusCodes->relaxNGValidate(__DIR__ . '/TestAsset/http-status-codes.rng')) {
+            self::fail();
```
weierophinney: This should have a useful message, to make it clear why the failure occurred (timeout reaching the validation file and/or invalid content in the file fetched). I can add that at merge, however.

test/ResponseTest.php
```diff
@@ -66,6 +66,63 @@ public function testReasonPhraseDefaultsToStandards()
         $this->assertEquals('Unprocessable Entity', $response->getReasonPhrase());
     }
 
+    public function ianaCodesReasonPhrasesProvider()
+    {
+        $ianaHttpStatusCodes = new \DOMDocument();
```
weierophinney: Import `DOMDocument`. (I can do that at merge)

test/ResponseTest.php
```diff
@@ -66,6 +66,63 @@ public function testReasonPhraseDefaultsToStandards()
...
+
+        $ianaHttpStatusCodes->load('https://www.iana.org/assignments/http-status-codes/http-status-codes.xml');
+
+        if (! $ianaHttpStatusCodes->relaxNGValidate(__DIR__ . '/TestAsset/http-status-codes.rng')) {
+            self::fail();
+        }
+
+        $ianaCodesReasonPhrases = [];
+
+        $xpath = new \DomXPath($ianaHttpStatusCodes);
```
weierophinney: Import `DOMXPath`. (I can do that at merge)

weierophinney:
Thanks, @fcabralpacheco

PR #235, opened 2017-02-22
Prefix authority with double slash if present.

ajgarlag:
The PSR-7 docblock of the `UriInterface::__toString` method, states that:

> If an authority is present, it MUST be prefixed by "//".

This PR tests and fixes that behavior.

### DIFF
--- a/test/UriTest.php
+++ b/test/UriTest.php
@@ -609,6 +609,12 @@ class UriTest extends TestCase
     public function testUriDoesNotAppendColonToHostIfPortIsEmpty()
     {
         $uri = (new Uri())->withHost('google.com');
-        $this->assertEquals('google.com', (string) $uri);
+        $this->assertEquals('//google.com', (string) $uri);
+    }
+
+    public function testAuthorityIsPrefixedByDoubleSlashIfPresent()
+    {
+        $uri = (new Uri())->withHost('example.com');
+        $this->assertEquals('//example.com', (string) $uri);
     }
 }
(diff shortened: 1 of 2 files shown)


test/UriTest.php
```diff
@@ -609,6 +609,12 @@ public function utf8QueryStringsDataProvider()
...
     {
         $uri = (new Uri())->withHost('google.com');
-        $this->assertEquals('google.com', (string) $uri);
+        $this->assertEquals('//google.com', (string) $uri);
+    }
+
+    public function testAuthorityIsPrefixedByDoubleSlashIfPresent()
+    {
+        $uri = (new Uri())->withHost('example.com');
+        $this->assertEquals('//example.com', (string) $uri);
```
asgrim: This test is exactly the same as `testUriDoesNotAppendColonToHostIfPortIsEmpty` except the URL used... not sure we need both except for being explicit.

ajgarlag: I was unsure too.

ajgarlag:
Every bug fixed is a sort of BC break as you cannot know if there are things that relay on the old behavior, but for me it should be fixed in order to be PSR7 compliant.

sharifzadesina:
@ajgarlag why you have closed this issue?

ajgarlag:
@sharifzadesina It was an error, 
is open again!

weierophinney:
Merged to develop for release with 1.4.0. Thanks, @ajgarlag

PR #240, opened 2017-03-29
Use the HTTP cookie header when available, instead of the  super global

DASPRiD:
PHP will replace special characters in cookie names, which results in other cookies not being available due to overwriting. Thus, the server request should take the cookies from the request header instead.

### DIFF
--- a/doc/book/usage.md
+++ b/doc/book/usage.md
@@ -87,6 +87,13 @@ $request = Zend\Diactoros\ServerRequestFactory::fromGlobals(
 );
 ```
 
+When no cookie array is supplied, `fromGlobals` will first try to parse the supplied `cookie` header
+before falling back to the `$_COOKIE` superglobal. This is done because PHP has some legacy handling
+for request parameters which were then registered as global variables. Due to this, cookies with a period
+in the name were renamed with underlines. By getting the cookies directly from the cookie header, you have
+access to the original cookies in the way you set them in your application and they are send by the user
+agent.
+
 ### Manipulating the response
 
 Use the response object to add headers and provide content for the response.  Writing to the body
--- a/src/ServerRequestFactory.php
+++ b/src/ServerRequestFactory.php
@@ -60,6 +60,10 @@ abstract class ServerRequestFactory
         $files   = static::normalizeFiles($files ?: $_FILES);
         $headers = static::marshalHeaders($server);
 
+        if (null === $cookies && array_key_exists('cookie', $headers)) {
+            $cookies = self::parseCookieHeader($headers['cookie']);
+        }
+
         return new ServerRequest(
             $server,
             $files,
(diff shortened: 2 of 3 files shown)


src/ServerRequestFactory.php
```diff
@@ -485,4 +489,34 @@ private static function marshalProtocolVersion(array $server)
 
         return $matches['version'];
     }
+
+    /**
+     * Parse a cookie header according to RFC 6265.
+     *
+     * PHP will replace special characters in cookie names, which results in other cookies not being available due to
+     * overwriting. Thus, the server request should take the cookies from the request header instead.
```
jeremiahsmall: We ran into a problem with interacting with a Crowd SSO cookie which is named `crowd.token_key` but was getting set in `$_COOKIE` as `crowd_token_key` and requiring special handling to ensure that those two variants of the string match. This PR would solve that problem.

doc/book/usage.md
```diff
@@ -87,6 +87,13 @@ $request = Zend\Diactoros\ServerRequestFactory::fromGlobals(
 );
 ```
 
+When no cookie array is supplied, `fromGlobals` will first try to parse the supplied `cookie` header
+before falling back to the `$_COOKIE` superglobal. This is done because PHP has some legacy handling
+for request parameters which were then registered as global variables. Due to this, cookies with a period
+in the name were renamed with underlines. By getting the cookies directly from the cookie header, you have
+access to the original cookies in the way you set them in your application and they are send by the user
+agent.
```
weierophinney: Question: what happens if `$_COOKIE` is passed? `fromGlobals()` allows passing an array of cookies, and some of our documentation actually demonstrates passing in the superglobals explicitly. Should we discourage that behavior when it comes to cookies?

If so, we should likely add a note to that effect, and change some of the examples.

DASPRiD: If `$_COOKIE` is passed there, you get the old behaviour. Not sure if we need to discourage it though. This is a rare case which this PR fixes, and usually people will call `fromGlobals()` without arguments I suppose?

jeremiahsmall: Even if we don't actively discourage, it's probably worth adding a note stating what the expected result would be so that the mentioned examples are consistent with the new feature. @weierophinney is there a good search term or other way to locate the examples you are thinking of?

weierophinney:
@sharifzadesina Putting a thumbs-down as a reaction without providing context as to _why_ is not terribly useful. Next time, please provide a comment if you disagree.

weierophinney:
Marking as milestone 2.7.0, as it's a feature change/addition, and could impact users who are not passing `$_COOKIE` when creating their request instance.

weierophinney:
Merged to develop for release with 1.4.0; thanks, @DASPRiD

PR #243, opened 2017-04-07
Pass the SAPI file upload errors when throwing an exception

Erikvv:
This will save people some time figuring out why the upload isn't working.

Can't think of any tests.

### DIFF
--- a/src/UploadedFile.php
+++ b/src/UploadedFile.php
@@ -17,15 +17,15 @@ use RuntimeException;
 class UploadedFile implements UploadedFileInterface
 {
     const ERROR_MESSAGES = [
-        UPLOAD_ERR_OK => 'There is no error, the file uploaded with success',
-        UPLOAD_ERR_INI_SIZE => 'The uploaded file exceeds the upload_max_filesize directive in php.ini',
-        UPLOAD_ERR_FORM_SIZE => 'The uploaded file exceeds the MAX_FILE_SIZE directive that was '
+        UPLOAD_ERR_OK         => 'There is no error, the file uploaded with success',
+        UPLOAD_ERR_INI_SIZE   => 'The uploaded file exceeds the upload_max_filesize directive in php.ini',
+        UPLOAD_ERR_FORM_SIZE  => 'The uploaded file exceeds the MAX_FILE_SIZE directive that was '
             . 'specified in the HTML form',
-        UPLOAD_ERR_PARTIAL => 'The uploaded file was only partially uploaded',
-        UPLOAD_ERR_NO_FILE => 'No file was uploaded',
+        UPLOAD_ERR_PARTIAL    => 'The uploaded file was only partially uploaded',
+        UPLOAD_ERR_NO_FILE    => 'No file was uploaded',
         UPLOAD_ERR_NO_TMP_DIR => 'Missing a temporary folder',
         UPLOAD_ERR_CANT_WRITE => 'Failed to write file to disk',
-        UPLOAD_ERR_EXTENSION => 'A PHP extension stopped the file upload.',
+        UPLOAD_ERR_EXTENSION  => 'A PHP extension stopped the file upload.',
     ];
 
     /**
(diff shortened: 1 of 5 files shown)


src/UploadedFile.php
```diff
@@ -114,7 +126,9 @@ public function __construct($streamOrFile, $size, $errorStatus, $clientFilename
     public function getStream()
     {
         if ($this->error !== UPLOAD_ERR_OK) {
-            throw new RuntimeException('Cannot retrieve stream due to upload error');
+            throw new RuntimeException(
+                'Cannot retrieve stream due to upload error: ' . self::$errorMessages[$this->error]
```
snapshotpl: Use `sprintf` to improve readability

src/UploadedFile.php
```diff
@@ -147,7 +161,9 @@ public function moveTo($targetPath)
         }
 
         if ($this->error !== UPLOAD_ERR_OK) {
-            throw new RuntimeException('Cannot retrieve stream due to upload error');
+            throw new RuntimeException(
+                'Cannot retrieve stream due to upload error: ' . self::$errorMessages[$this->error]
```
snapshotpl: What if `$this->error` contains unsupported (in `errorMessages`) value?

src/UploadedFile.php
```diff
@@ -16,6 +16,18 @@
 
 class UploadedFile implements UploadedFileInterface
 {
+    private static $errorMessages = [
```
snapshotpl: Is `static` necessary?

src/UploadedFile.php
Erikvv: This invariant is already enforced in the constructor
```
        if (! is_int($errorStatus)
            || 0 > $errorStatus
            || 8 < $errorStatus
        ) {
            throw new InvalidArgumentException(
                'Invalid error status for UploadedFile; must be an UPLOAD_ERR_* constant'
            );
        }
```

src/UploadedFile.php
Erikvv: Done

src/UploadedFile.php
Erikvv: This is a good use case because it is immutable data. If it wasn't static it would be duplicated and it would increase the memory footprint by a lot.

Erikvv:
Can anyone please merge this?

src/UploadedFile.php
weierophinney: Make this a constant instead; PHP has supported constant array values since 5.6, and they're a better fit here than a static member.

src/UploadedFile.php
```diff
@@ -114,7 +126,10 @@ public function __construct($streamOrFile, $size, $errorStatus, $clientFilename
     public function getStream()
     {
         if ($this->error !== UPLOAD_ERR_OK) {
-            throw new RuntimeException('Cannot retrieve stream due to upload error');
+            throw new RuntimeException(sprintf(
+                'Cannot retrieve stream due to upload error: %s',
+                self::$errorMessages[$this->error]
+            ));
```
weierophinney: This can be tested by passing a non-`UPLOAD_ERR_OK` value to the `$errorStatus` constructor argument, and then calling this method.

src/UploadedFile.php
```diff
@@ -147,7 +162,10 @@ public function moveTo($targetPath)
         }
 
         if ($this->error !== UPLOAD_ERR_OK) {
-            throw new RuntimeException('Cannot retrieve stream due to upload error');
+            throw new RuntimeException(sprintf(
+                'Cannot retrieve stream due to upload error: %s',
+                self::$errorMessages[$this->error]
+            ));
```
weierophinney: Same comment here.

weierophinney:
I've taken the liberty of making the requested changes and pushing them to your branch, @Erikvv .

weierophinney:
Thanks, @Erikvv

PR #247, opened 2017-04-18
__toString() always return empty content when resource is not seekable

svycka:
I have noticed this with amazon S3 it does return resource with meta `seekable => false` and I can't use this I have to create file and the n create new stream from it. Also does not make sense to rewind not seekable stream.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ cache:
     - $HOME/.local
     - zf-mkdoc-theme
 
+dist: trusty
+
 env:
   global:
     - COMPOSER_ARGS="--no-interaction"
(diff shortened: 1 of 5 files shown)


svycka:
what's strange here that hhvm did not fail here: https://travis-ci.org/zendframework/zend-diactoros/builds/223063138
@weierophinney  also not sure but maybe we should add this check also for phpStream and RelativeStream?

weierophinney:
@svycka The `PhpInputStream` is always seekable, and we have functionality to ensure rewinding works (by caching the data as we read). The `RelativeStream` implementation, since it delegates to another stream, likely needs the change as well; I'll try and write a test as I merge this to see what happens.

weierophinney:
Thanks, @svycka

PR #248, opened 2017-05-10
Encode user info when calling withUserInfo()

DASPRiD:
According to <https://www.ietf.org/rfc/rfc3986.txt> 3.2.1, the user info parts should actually be URL encoded.

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -166,6 +166,10 @@ class Uri implements UriInterface
     }
 
     /**
+     * Retrieve the user-info part of the URI.
+     *
+     * This value is percent-encoded, per RFC 3986 Section 3.2.1.
+     *
      * {@inheritdoc}
      */
     public function getUserInfo()
@@ -242,6 +246,11 @@ class Uri implements UriInterface
     }
 
     /**
+     * Create and return a new instance containing the provided user credentials.
+     *
+     * The value will be percent-encoded in the new instance, but with measures
+     * taken to prevent double-encoding.
+     *
      * {@inheritdoc}
      */
     public function withUserInfo($user, $password = null)
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -443,7 +443,7 @@ class Uri implements UriInterface
         }
 
         $this->scheme    = isset($parts['scheme']) ? $this->filterScheme($parts['scheme']) : '';
-        $this->userInfo  = isset($parts['user']) ? $parts['user'] : '';
+        $this->userInfo  = isset($parts['user']) ? $this->filterUserInfoPart($parts['user']) : '';
         $this->host      = isset($parts['host']) ? $parts['host'] : '';
         $this->port      = isset($parts['port']) ? $parts['port'] : null;
         $this->path      = isset($parts['path']) ? $this->filterPath($parts['path']) : '';
@@ -555,8 +555,10 @@ class Uri implements UriInterface
      */
     private function filterUserInfoPart($part)
     {
+        // Note the addition of `%` to initial charset; this allows `|` portion
+        // to match and thus prevent double-encoding.
         return preg_replace_callback(
-            '/(?:[^' . self::CHAR_UNRESERVED . self::CHAR_SUB_DELIMS . ']+|%(?![A-Fa-f0-9]{2}))/u',
+            '/(?:[^%' . self::CHAR_UNRESERVED . self::CHAR_SUB_DELIMS . ']+|%(?![A-Fa-f0-9]{2}))/u',
             [$this, 'urlEncodeChar'],
             $part
         );
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -25,14 +25,14 @@ use Psr\Http\Message\UriInterface;
 class Uri implements UriInterface
 {
     /**
-     * Sub-delimiters used in query strings and fragments.
+     * Sub-delimiters used in user info, query strings and fragments.
      *
      * @const string
      */
     const CHAR_SUB_DELIMS = '!\$&\'\(\)\*\+,;=';
 
     /**
-     * Unreserved characters used in paths, query strings, and fragments.
+     * Unreserved characters used in user info, paths, query strings, and fragments.
      *
      * @const string
      */
@@ -261,9 +261,9 @@ class Uri implements UriInterface
             ));
         }
 
-        $info = $user;
+        $info = $this->filterUserInfoPart($user);
         if ($password) {
-            $info .= ':' . $password;
+            $info .= ':' . $this->filterUserInfoPart($password);
         }
 
         if ($info === $this->userInfo) {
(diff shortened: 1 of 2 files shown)


src/Uri.php
```diff
@@ -261,9 +261,9 @@ public function withUserInfo($user, $password = null)
             ));
         }
 
-        $info = $user;
+        $info = $this->filterUserInfoPart($user);
         if ($password) {
-            $info .= ':' . $password;
+            $info .= ':' . $this->filterUserInfoPart($password);
```
jeremiahsmall: This patch will fix most cases, however, if the password contains an apparently url encoded string literal, for example, "%3A" that will be decoded as ":" and therefore fail authentication. In discussion with @DASPRiD, fixing this would be a BC break.

jeremiahsmall: On second thought, maybe we can change this to a BC compatible solution by _not_ using `self::CHAR_UNRESERVED` and `self::CHAR_SUB_DELIMS` for user info. Instead, we can/should/must assume that input is already urlencoded if it comes from the uri string, and can/should/must assume it's not if it is set via `withUserInfo` method.

```
        $info = urlencode($user);
        if ($password) {
            $info .= ':' . urlencode($password);
        }
```

Now, if that would be considered bc breaking, we could add a third optional arg to `withUserInfo`. E.g.

```
public function withUserInfo($user, $password = null, $urlEncoded = true)
    {
   // snip
```

But I'd argue that if someone is already urlencoding this input, they're working around the bug, and we shouldn't worry about BC.

src/Uri.php
```diff
@@ -548,6 +548,21 @@ private function filterScheme($scheme)
     }
 
     /**
+     * Filters a part of user info in a URI to ensure it is properly encoded.
+     *
+     * @param string $part
+     * @return string
+     */
+    private function filterUserInfoPart($part)
```
jeremiahsmall: @weierophinney, see my prior comment for argument in favor of using SPL `urlencode` function instead of this.

src/Uri.php
jeremiahsmall: In discussion with @DASPRiD we realized that since the PSR-7 interface doesn't include the third argument, that suggests we shouldn't add it in Diactoros, even though PSR-7 does not specify if the input should be url encoded or not.

Also, @DASPRiD argues that using the SPL urlencode is overkill because it encodes too many characters, but agrees that always encoding for the user info inputs is correct since it would be a pretty normal case for a password to contain a character combination that looks like an encoded character, but is in fact a string literal.

test/UriTest.php
```diff
@@ -87,6 +87,14 @@ public function testWithUserInfoReturnsNewInstanceIfUserAndPasswordAreSameAsBefo
         $this->assertEquals('https://user:pass@local.example.com:3001/foo?bar=baz#quz', (string) $new);
     }
 
+    public function testWithUserInfoEncodesUsernameAndPassword()
+    {
+        $uri = new Uri('https://user:pass@local.example.com:3001/foo?bar=baz#quz');
+        $new = $uri->withUserInfo('foo:bar', 'baz:bat');
+
+        $this->assertSame('foo%3Abar:baz%3Abat', $new->getUserInfo());
+    }
```
jeremiahsmall: Probably either a second assertion, or add `%25` to the first input to cover that it's not getting accepted as urlencoded input.

```
public function testWithUserInfoDoesNotExpectUrlencodedCharactersInUsernameAndPassword()
{
    $uri = new Uri('https://localhost');
    $new = $uri->withUserInfo('%25', '%25');

   $this->assertSame('%2525:%2525', $new->getUserInfo());
}
```

weierophinney:
I've created a new commit that adds a data provider with both user and credential values to test against, as well as the expected value of `getUserInfo()` following, and pushed that; it covers the suggested test by @jeremiahsmall as well.

The one issue I see with this is that it can lead to issues if you use the value retrieved from `getUserInfo()` to seed a new instance (e.g., `$instance = $instance->withUserInfo($previous->getUserInfo())`; you'll essentially get double-encoding (as evidenced by the test @jeremiahsmall suggested, where `%25` becomes `%2525`).

Interestingly, @akrabat ran into this issue this week as well, when using an email address as the user information (e.g., `user@example.com`).

My suggestion is that any encoding is done when calling `getAuthority()` (which is used internally by `createUriString()` to return the authority value, which includes user-info, host, and port); this allows passing the raw values to `withUserInfo()`, and for `getUserInfo()` to return the same data it receives.

In reviewing the code, I also see that the `parseUri()` implementation is buggy, because it does not `urldecode()` the `user` and `pass` segments, which it likely should to prevent issues when consuming these values. (Interestingly, PHP's `parse_url()` function will accept unencoded values here and still parse the URL correctly!)

To be honest, `parseUri()` likely should not filter the path, query, and fragment, and instead delegate those operations to `createUriString()`. However, I think that's a topic for another PR.

@akrabat @DASPRiD and @jeremiahsmall : Do my suggestions here make sense?

jeremiahsmall:
The new data provider in the test looks good. Not sure what to make of the suggestion about encoding when calling `getAuthority()` without more time to reflect on that.

weierophinney:
I'm going to move this to the 1.5.0 milestone, so we can also address the other issues related to `parseUri()`; I think consistency in how we handle these is best, and I'd rather the various getter methods return the raw values, and that casting to string filters/normalizes them for that context. This prevents double-encoding issues

The only "con" I can see is that doing so puts the onus on the user to encode if they wish to use those values as part of a URI string). I think we can message that in the docblocks, however.

jeremiahsmall:
So are we still saying that the arguments in `withUserInfo()` would be unencoded. If they would need to be encoded by the user, the variable names should reflect that. Eg:


```
public function withUserInfo($user, $password = null)
    {
   // snip
```
vs
```
public function withUserInfo($urlEncodedUser, $urlEncodedPassword = null)
    {
   // snip
```

weierophinney:
If they're not encoded, why would we indicate they're encoded in the variable name?

jeremiahsmall:
Right, well I guess that answers my question... I was making sure that you're not suggesting that these be encoded.

weierophinney:
> I'm going to move this to the 1.5.0 milestone, so we can also address the other issues related to parseUri(); I think consistency in how we handle these is best, and I'd rather the various getter methods return the raw values, and that casting to string filters/normalizes them for that context. This prevents double-encoding issues

Okay, re-read the specification, and this is not possible. Each of the docblocks for `getPath()`, `getQuery()`, and `getFragment()` in the `UriInterface` indicate that they should return the _percent encoded values_, and prevent double-encoding. As such, the current approach suggested by @DASPRiD in this pull request is correct: percent-encode the value as it comes in, and any call to `getUserInfo()` and/or `getAuthority()` should return that encoded value. This will keep the workflow consistent with the other URI segments.

What this means, though, is that we need to ensure that the `filterUserInfoPart()` does not double-encode. We have tests in place for the other URI segments that use encoding, so I can use those as a basis for understanding how to both test and implement that functionality.

weierophinney:
@jeremiahsmall I've isolated the double-encoding issue, and have now pushed a test that verifies that calling `filterUserInfoPart()` no longer double-encodes. With that change, and assuming that the changes pass continuous integration, I can merge this finally.

weierophinney:
Thanks, @DASPRiD and @jeremiahsmall

jeremiahsmall:
+1

PR #249, opened 2017-05-10
Do not create a new instance when nothing changed

DASPRiD:
When not changing an object through one of the immutable methods, there is no need to return a cloned instance with the exact same properties.


weierophinney:
This is a slight (and minor) behavior change, so scheduling for 1.5.0.

weierophinney:
Thanks, @DASPRiD

weierophinney:
Merged to develop for release in 1.5.0.

PR #257, opened 2017-07-13
fix #255 - Don't evaluate '0' as empty

### LINKED ISSUE #255: PhpInputStream and '0' chunks
`PhpInputStream::read` doesn't add `"0"` chunks to `$this->cache`, because the [first condition of the `if`](https://github.com/zendframework/zend-diactoros/blob/fbcf611e8766a599c4839da569f5e2d267ddf201/src/PhpInputStream.php#L62) evaluates to `false`.

sasezaki:
Psr-7 StreamInterface::read() returns only string, so don't need check it is empty-like value
fix #255

### DIFF
--- a/test/PhpInputStreamTest.php
+++ b/test/PhpInputStreamTest.php
@@ -64,6 +64,17 @@ class PhpInputStreamTest extends TestCase
         $this->assertEquals(substr($contents, 128), $remainder);
     }
 
+    public function testGetContentsReturnCacheWhenReachedEof()
+    {
+        $this->stream->getContents();
+        $this->assertStreamContents($this->stream->getContents());
+
+        $stream = new PhpInputStream('data://,0');
+        $stream->read(1);
+        $stream->read(1);
+        $this->assertSame('0', $stream->getContents(), 'Don\'t evaluate 0 as empty');
+    }
+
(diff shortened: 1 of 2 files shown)


test/PhpInputStreamTest.php
```diff
@@ -64,6 +64,17 @@ public function testGetContentsReturnsRemainingContentsOfStream()
         $this->assertEquals(substr($contents, 128), $remainder);
     }
 
+    public function testGetContentsReturnCacheWhenReachedEof()
+    {
+        $this->stream->getContents();
+        $this->assertStreamContents($this->stream->getContents());
+
+        $stream = new PhpInputStream('data://,0');
```
kelunik: What does `data://,0` mean?

sasezaki: I intended just represent `string(1) "0"` for PHP's stream. ( http://php.net/wrappers.data )

bwoebi: a very trivial data:// URL, without mime, charset or base64. (see RFC 2397, 3. syntax)
In this case, with just `0` as contents.

weierophinney:
Thanks, @sasezaki

PR #258, opened 2017-07-15
reserved characters in path should be unencoded

samsonasik:
like in zend-uri issue https://github.com/zendframework/zend-uri/issues/13 with zend-uri PR https://github.com/zendframework/zend-uri/pull/17

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -556,7 +556,7 @@ class Uri implements UriInterface
     private function filterPath($path)
     {
         $path = preg_replace_callback(
-            '/(?:[^' . self::CHAR_UNRESERVED . ':@&=\+\$,\/;%]+|%(?![A-Fa-f0-9]{2}))/u',
+            '/(?:[^(' . self::CHAR_UNRESERVED . '):@&=\+\$,\/;%]+|%(?![A-Fa-f0-9]{2}))/u',
             [$this, 'urlEncodeChar'],
             $path
         );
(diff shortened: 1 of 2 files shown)


samsonasik:
if the "un" prefix usage in the pr/tests for "unencoded" or "unreserved" words are not grammatically correct, please let me know so I can change them.

src/Uri.php
```diff
@@ -556,7 +556,7 @@ private function filterScheme($scheme)
     private function filterPath($path)
     {
         $path = preg_replace_callback(
-            '/(?:[^' . self::CHAR_UNRESERVED . ':@&=\+\$,\/;%]+|%(?![A-Fa-f0-9]{2}))/u',
+            '/(?:[^(' . self::CHAR_UNRESERVED . '):@&=\+\$,\/;%]+|%(?![A-Fa-f0-9]{2}))/u',
```
weierophinney: This is confusing the way you have it written; it looks like a mistake, though it's not.

I thought at first you were trying to create a capturing group by placing parens, and, since this expression is within square brackets, that would not work.

What you *are* trying to do is add parens to the character list, and, as such, *it's not a mistake*. It just *reads like one*.

As such, when I merge, I'm going to disambiguate and push them both into the concatenated string. Further, one common practice when writing a character group for a regex is to flip the order of parens, braces, etc: `)(` instead of `()`. This helps ensure that you don't accidentally create a capturing group, a character group, or a quantifying expression.

weierophinney:
Thanks, @samsonasik

samsonasik:
@weierophinney Thank you, I've updated the zend-uri PR at https://github.com/zendframework/zend-uri/pull/17 to use `)(` instead of `()`

PR #259, opened 2017-07-15
dist: trusty for hhvm for travis

samsonasik:
due to travis error: 
```
HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with `dist: trusty`.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: false
-
+dist: trusty
 language: php
 
 cache:


weierophinney:
We're dropping hhvm support; will this benefit other builds?

samsonasik:
not sure, it is just in case we still need it as it listed in travis config, I will remove the hhvm config in other PR

samsonasik:
@weierophinney I've created PR https://github.com/zendframework/zend-diactoros/pull/260 for removing hhvm support

webimpress:
It can be closed now, trusty will be default dist soon (are even it is already default one).

PR #267, opened 2017-08-24
Small optimizations

snapshotpl:


src/CallbackStream.php
```diff
@@ -172,10 +172,10 @@ public function getMetadata($key = null)
             return $metadata;
         }
 
-        if (! array_key_exists($key, $metadata)) {
-            return null;
+        if (isset($metadata[$key])) {
```
Ocramius: `return $metadata[$key] ?? null;`

src/RelativeStream.php
```diff
@@ -107,7 +107,7 @@ public function isSeekable()
      */
     public function seek($offset, $whence = SEEK_SET)
     {
-        if ($whence == SEEK_SET) {
+        if ($whence === SEEK_SET) {
```
Ocramius: Can you use the FQN for this constant?

src/ServerRequestFactory.php
```diff
@@ -60,7 +60,7 @@ public static function fromGlobals(
         $files   = static::normalizeFiles($files ?: $_FILES);
         $headers = static::marshalHeaders($server);
 
-        if (null === $cookies && array_key_exists('cookie', $headers)) {
+        if (null === $cookies && isset($headers['cookie'])) {
```
Ocramius: You need to introduce tests around `null` values to avoid regressions here

src/ServerRequestFactory.php
```diff
@@ -206,7 +206,7 @@ public static function marshalHeaders(array $server)
 
                 // We will not overwrite existing variables with the
                 // prefixed versions, though
-                if (array_key_exists($key, $server)) {
+                if (isset($server[$key])) {
                     continue;
```
Ocramius: You need to introduce tests around `null` values to avoid regressions here

src/Stream.php
```diff
@@ -285,11 +285,11 @@ public function getMetadata($key = null)
         }
 
         $metadata = stream_get_meta_data($this->resource);
-        if (! array_key_exists($key, $metadata)) {
-            return null;
+        if (isset($metadata[$key])) {
```
Ocramius: Use `??` here too

src/Uri.php
```diff
@@ -641,7 +641,7 @@ private function filterQuery($query)
     private function splitQueryValue($value)
     {
         $data = explode('=', $value, 2);
-        if (1 === count($data)) {
+        if (! isset($data[1])) {
```
Ocramius: This check is not equivalent to the previous: you need to verify that the key `0` exists too

src/CallbackStream.php
snapshotpl: Good idea but after drop php5 ;(

src/RelativeStream.php
snapshotpl: I'm not sure what are you meant

Ocramius: `\SEEK_SET`

src/Uri.php
samsonasik: @Ocramius `explode()` should always has key 0 ? https://3v4l.org/viNZT

src/ServerRequestFactory.php
snapshotpl: I think it's imposible to write this test becase `$headers` comes from private function which alware return strings pair array.

src/Uri.php
snapshotpl: @Ocramius agree with @samsonasik

src/CallbackStream.php
Ocramius: Wha?! This is for an improvement: just drop it already

src/ServerRequestFactory.php
Ocramius: No need then: nullability was just an assumption on my end.

src/Uri.php
Ocramius: Yup, misread

src/CallbackStream.php
snapshotpl: But is it not better to merge it as it to current php5 supported line and then, drop php5 and apply you have suggested?

snapshotpl:
@Ocramius all remarks applied in code or I miss something?

webimpress:
@snapshotpl tests are failing. Why?

src/RelativeStream.php
```diff
@@ -105,9 +103,9 @@ public function isSeekable()
     /**
      * {@inheritdoc}
      */
-    public function seek($offset, $whence = SEEK_SET)
+    public function seek($offset, $whence = \SEEK_SET)
     {
-        if ($whence == SEEK_SET) {
+        if ($whence === \SEEK_SET) {
```
webimpress: I would go with import the constant as we do in other places for internal functions:
```php
use const SEEK_SET;
```

src/ServerRequestFactory.php
weierophinney: Agreed, particularly as the values may resolve to a `false` (e.g., `0`, `''`). The reason `array_key_exists()` was used was to allow _any_ value, regardless of whether or not it evaluated as truthy.

src/Stream.php
```diff
@@ -193,11 +195,11 @@ public function isWritable()
...
-            strstr($mode, 'x')
-            || strstr($mode, 'w')
-            || strstr($mode, 'c')
-            || strstr($mode, 'a')
-            || strstr($mode, '+')
+            strpos($mode, 'x') !== false
+            || strpos($mode, 'w') !== false
+            || strpos($mode, 'c') !== false
+            || strpos($mode, 'a') !== false
+            || strpos($mode, '+') !== false
```
weierophinney: How is this change "better" or an "optimization", exactly? I find it less readable.

src/Stream.php
```diff
@@ -234,7 +236,7 @@ public function isReadable()
         $meta = stream_get_meta_data($this->resource);
         $mode = $meta['mode'];
 
-        return (strstr($mode, 'r') || strstr($mode, '+'));
+        return (strpos($mode, 'r') !== false || strpos($mode, '+') !== false);
```
weierophinney: Same comment as above regarding `strstr` vs `strpos`. I recognize that `strpos` is marginally faster, but most folks agree it's a micro-optimization at best. If using it sacrifices readability, I'm not sure the change is useful.

src/Stream.php
weierophinney: If you target PHP 7.1+, you'll need to update the `composer.json`, rebase against develop, and change the branch this is against.

src/CallbackStream.php
weierophinney: If we can optimize for PHP 5.6 today, we can make it even better for the next _major_ release later. I'm fine with this for now.

weierophinney:
Closing; many changes were requested, but none provided by the author.

If you or anyone else reading want to see these changes, please submit a new PR incorporating them, along with the requested feedback.

Thanks!

PR #269, opened 2017-09-05
Hotfix - Travis CI configuration - removed double coverage run

webimpress:
Removed double test coverage run - run it once only on PHP 7.
Simplified condition to remove xdebug when not running tests with coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,8 +56,8 @@ matrix:
     - php: 7.2
 
 before_install:
+  - if [[ $TEST_COVERAGE != 'true' ]]; then phpenv config-rm xdebug.ini || return 0 ; fi
   - travis_retry composer self-update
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -49,16 +49,15 @@ matrix:
...
     - php: 7.2
       env:
         - DEPS=latest
   allow_failures:
     - php: 7.2
 
 before_install:
+  - if [[ $TEST_COVERAGE != 'true' ]]; then phpenv config-rm xdebug.ini || return 0 ; fi
   - travis_retry composer self-update
-  - if [[ $TEST_COVERAGE != 'true' && "$(php --version | grep xdebug -ci)" -ge 1 ]]; then phpenv config-rm xdebug.ini ; fi
```
weierophinney: The check to see if xdebug is present should still occur, as we cannot be certain php 7.2 works with it yet.

I'll do that during merge, however.

weierophinney:
Thanks, @webimpress

PR #273, opened 2017-10-10
Respect response's status code

lcobucci:
This basically ensures that `zend-diactoros` sends the status code to `header()` while emitting the response headers, preventing PHP from overriding it silently.

IMO this is a bug fix, since the emitted response is different from the created response.

We have a similar solution in `symfony/http-foundation`: https://github.com/symfony/http-foundation/blob/dd97248b62153457d0716738d0dbcc0c2b31e285/Response.php#L331-L338

More info:

- https://bugs.php.net/bug.php?id=70273
- https://bugs.php.net/bug.php?id=51749
- https://bugs.php.net/bug.php?id=74535

### DIFF
--- a/src/Response/SapiEmitter.php
+++ b/src/Response/SapiEmitter.php
@@ -26,8 +26,8 @@ class SapiEmitter implements EmitterInterface
     {
         $this->assertNoPreviousOutput();
 
-        $this->emitStatusLine($response);
         $this->emitHeaders($response);
+        $this->emitStatusLine($response);
         $this->emitBody($response);
     }
 
--- a/src/Response/SapiEmitterTrait.php
+++ b/src/Response/SapiEmitterTrait.php
@@ -38,7 +38,13 @@ trait SapiEmitterTrait
      * Emits the status line using the protocol version and status code from
      * the response; if a reason phrase is available, it, too, is emitted.
      *
+     * It's important to mention that, in order to prevent PHP from changing
+     * the status code of the emitted response, this method should be called
+     * after `emitHeaders()`
+     *
      * @param ResponseInterface $response
+     *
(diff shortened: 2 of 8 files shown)


weierophinney:
In looking at this, I think there may be a better approach.

Each of the two SAPI emitter implementations call the following:

```php
$this->emitStatusLine($response);
$this->emitHeaders($response);
```

I'm thinking that perhaps those lines should be reversed, and `emitStatusLine()` modified to pass the `$http_response_code` argument when calling `header()`.

Alternately, we keep your suggested solution, but _also_ make those changes, to be 100% certain in situations where custom implementations call the methods in a different order.

Thoughts?

lcobucci:
@weierophinney reversing the calls also fixes it indeed. However IMO is a bit fragile, specially having the possibility of creating custom emitters that shares the trait, and not really "natural" - I expect that the header stack would be status line and then the headers.

I'd suggest to only force the response code when calling `header()`, but I can modify this PR according to your preference.

weierophinney:
> I'd suggest to only force the response code when calling header(), but I can modify this PR according to your preference.

Well, `emitStatusLine()` also calls `header()`, so we should do it there, too.

Let's do the following:

- Add the response code when calling `header()` in `emitStatusLine()`
- Reverse the `emitStatusLine()` and `emitHeaders()` calls in both implementations
- Add information in the `emitStatusLine()` docblock indicating that method should generally be called _after_ `emitHeaders()`.

This will cover all bases, I think.

Thanks, @lcobucci !

lcobucci:
@weierophinney done, thanks!

weierophinney:
Thanks, @lcobucci! Merged, and released with 1.6.1.

PR #275, opened 2017-10-21
Nits on range check for port number, unnecessary sprintf

TysonAndre:
The old version would be parsed as `($port !== null && $port < 1) || ...`

(That'd have the same results for `null` and `0`,
but is less readable)


weierophinney:
Thanks, @TysonAndre

PR #280, opened 2017-12-04
Add early hint status code

snapshotpl:
https://datatracker.ietf.org/doc/draft-ietf-httpbis-early-hints/


weierophinney:
Thanks, @snapshotpl!

PR #282, opened 2017-12-14
Updated to php-coveralls/php-coveralls

webimpress:
With version 2 package has been renamed from `satooshi/php-coveralls` to `php-coveralls/php-coveralls`, and the script has been renamed from `coveralls` to `php-coveralls`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,6 @@
       "@cs-check",
       "@test"
     ],
-    "upload-coverage": "coveralls -v",
     "cs-check": "phpcs",
     "cs-fix": "phpcbf",
     "test": "phpunit --colors=always",
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ cache:
 env:
   global:
     - COMPOSER_ARGS="--no-interaction"
-    - COVERAGE_DEPS="satooshi/php-coveralls"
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -54,7 +54,6 @@
       "@cs-check",
       "@test"
     ],
-    "upload-coverage": "coveralls -v",
```
weierophinney: Why this change, exactly? The point of having a script is so that we can make a change to the script, without having to make a change to the travis configuration...

This should change to:

```json
"upload-coverage": "php-coveralls -v",
```

and then the `.travis.yml` only needs a change in terms of the `COVERAGE_DEPS` declaration.

I'll take care of that during merge.

weierophinney:
Thanks, @webimpress.

composer.json
webimpress: @weierophinney I think this is what we agreed on slack and what was merged into maintainers repository:
https://github.com/zendframework/maintainers/pull/38

The script is only used in travis build and imho it shouldn't be available to the developers.

weierophinney:
@Xerkus We do not need to add patches that only change docs or QA tooling to milestones or to the changelogs; these are internal and not user facing.

composer.json
weierophinney: @webimpress I'd totally forgotten that conversation!

I've opened #288 to re-instate these changes.

PR #283, opened 2017-12-28
Make it possible to override fromGlobals easily

sasanrose:
If I wanted to use a custom ServerRequest which extends Zend diactoros ServerRequest and still use the ServerRequestFactory::fromGlobals it was not possible without duplicating some methods and code. This change make it easier to extend ServerRequestFactory, override fromGlobals method and as a result use your own custom version of ServerRequest.

### DIFF
--- a/src/ServerRequestFactory.php
+++ b/src/ServerRequestFactory.php
@@ -54,26 +54,64 @@ abstract class ServerRequestFactory
         array $cookies = null,
         array $files = null
     ) {
-        $server  = static::normalizeServer($server ?: $_SERVER);
-        $files   = static::normalizeFiles($files ?: $_FILES);
-        $headers = static::marshalHeaders($server);
-
-        if (null === $cookies && array_key_exists('cookie', $headers)) {
-            $cookies = self::parseCookieHeader($headers['cookie']);
-        }
+        list(
+            $server,
+            $cookies,
+            $files,
+            $headers,
+            $uri,
+            $method,
+            $protocolVersion
+        ) = static::prepareGlobals($server, $cookies, $files);
 
         return new ServerRequest(
             $server,
             $files,
-            static::marshalUriFromServer($server, $headers),
-            static::get('REQUEST_METHOD', $server, 'GET'),
+            $uri,
+            $method,
             'php://input',
             $headers,
             $cookies ?: $_COOKIE,
             $query ?: $_GET,
             $body ?: $_POST,
-            static::marshalProtocolVersion($server)
+            $protocolVersion
         );
+    }
+
+    /**
+     * Prepare required variables to create and instance using globals.
+     *
+     * @param array $server $_SERVER superglobal
+     * @param array $cookies $_COOKIE superglobal
+     * @param array $files $_FILES superglobal
+     *
+     * @return array
+     */
+    protected static function prepareGlobals(
+        array $server = null,
+        array $cookies = null,
+        array $files = null
+    ) {
+        $server          = static::normalizeServer($server ?: $_SERVER);
+        $files           = static::normalizeFiles($files ?: $_FILES);
+        $headers         = static::marshalHeaders($server);
+        $uri             = static::marshalUriFromServer($server, $headers);
+        $method          = static::get('REQUEST_METHOD', $server, 'GET');
+        $protocolVersion = static::marshalProtocolVersion($server);
+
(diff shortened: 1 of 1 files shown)


sasanrose:
A test is failing in master branch which fails in my branch as well. I did not know how to fix it properly. The test is not related to this pull request.

src/ServerRequestFactory.php
```diff
@@ -472,7 +510,7 @@ private static function normalizeNestedFileSpec(array $files = [])
      * @param array $server
      * @return string
      */
-    private static function marshalProtocolVersion(array $server)
+    public static function marshalProtocolVersion(array $server)
```
Ocramius: To be reverted - this is `private` API for which we shouldn't provide BC guarantees

src/ServerRequestFactory.php
```diff
@@ -497,7 +535,7 @@ private static function marshalProtocolVersion(array $server)
      * @param $cookieHeader
      * @return array
      */
-    private static function parseCookieHeader($cookieHeader)
+    public static function parseCookieHeader($cookieHeader)
```
Ocramius: To be reverted - this is `private` API for which we shouldn't provide BC guarantees

Ocramius:
Not closing yet: need an opinion from @weierophinney.

I don't think C&P is as bad as it sounds, compared to the introduced BC boundaries.

weierophinney:
I tend to agree with @Ocramius here; the factory provided is specific to our `ServerRequest` implementation, and if you want to change how it does its work, you should create your own implementation, and/or provide a pull request with the changes and an argument for why we should alter how we currently do the work.

The various aspects of this particular factory are the result of our experiences in managing incoming requests since the early days of ZF1 through the present, and have been vetted a ton. Making it possible to extend the factory in order to alter that behavior will lead to more maintenance issues for us down the line. Forcing users to create their own factories in those situations will force users to fully consider the potential impact of the changes they want to introduce.

sasanrose:
@weierophinney @Ocramius Sorry for late reply. I see that you guys closed this request and I understand your concerns. But I believe one of the strengths of zend-diactoros is the ServerRequestFactory class and all it's useful methods. A lot of them can be changed to be reused without even having any effects on the performance or behaviour.
Imagine I have a ServerRequest that goes through a bunch of middlewares and then passed to my controllers. In one of my middlewares I want to parse some query parameters and store it in my ServerRequest (i.e withKeys or withAction) and then receive them in my controllers (getKeys or getActions). Just for this I have to go through the re-implementing or duplicating a lot of ServerRequestFactory methods. While I could reuse it easily.

Ocramius:
> A lot of them can be changed to be reused without even having any effects on the performance or behaviour.

The problems are:

 * stability
 * BC boundaries
 * coupling

Inheritance is the worst kind of coupling after shared singleton state.

 > Imagine I have a ServerRequest that goes through a bunch of middlewares and then passed to my controllers. In one of my middlewares I want to parse some query parameters and store it in my ServerRequest (i.e withKeys or withAction) and then receive them in my controllers (getKeys or getActions). Just for this I have to go through the re-implementing or duplicating a lot of ServerRequestFactory methods. While I could reuse it easily.

You are probably looking for `withAttribute`. The correct approach to this is to use that, even if it has less type safety. You can see examples of how that is correctly used in https://github.com/psr7-sessions/storageless/blob/e9d5b23503e68b10d19ac637731fef86ebb9f4ca/src/Storageless/Http/SessionMiddleware.php#L184 or https://github.com/dflydev/dflydev-fig-cookies.

weierophinney:
@sasanrose For a reference, here's an excellent blog post detailing some of the issues we have with modifying classes to open the API: https://medium.com/@nicolopigna/inheritance-is-evil-stop-using-it-6c4f1caf5117

Essentially, what happens if we change the behavior of one of the methods you propose changing from `private` to `protected`? What if you depended on the previous behavior? We've now broken _your_ code by making a change in _our_ code.

Yes, this means more work for _you_ if you want to extend the class and use the existing behavior. However, it also isolates your implementation from upstream changes, ensuring it works consistently for your own needs. If we report bugfixes that you are interested in, you can apply the same patches to your code; however, if those bugfixes would represent a BC break for you, you're free to ignore them.

sasanrose:
@weierophinney Fair enough. Thanks for the explanation and the reference

Issue #189, opened 2016-07-15
Add XmlResponse


xorock:
What do You think about adding one more response type for xml documents? With default header application/xml.

weierophinney:
Fixed with #285.

PR #285, opened 2018-01-03
Add simple XML Response Ref #189

gsteel:


Xerkus:
Why do we have those at all? Those extra response objects do not bring any value. Any dependency must be on interface, so they can't be used for typehinting. Whatever is creating new XML response, or Json response, could just set content type itself, it already knows enough to do so.

Xerkus:
Test failure is unrelated. IANA published new status code

gsteel:
Fair enough, but the convenience of not having to repeat yourself whenever you want to return an HTML/Json/Text/XML response is helpful no?

Xerkus:
I suppose I was too hasty to comment. It certainly saves couple lines of code.
You see, there is a PSR-17, HTTP factory interface, that enables consumers not to depend on diactoros at all. On the other hand this little convenience does not really affect interoperability. I can live with it.

weierophinney:
@Xerkus these convenience response types are primarily for application developers, who are producing request handlers/inner middleware, and who already know the PSR-7 implementation they are choosing. Having these is a differentiator with other implementations, and makes producing common response types simpler. The factories in PSR-17, while generally useful, still make some things convoluted (e.g., producing a JSON response, an HTML or XML response, a redirect, etc.).

PSR-17, in my mind, is more around writing _distributable middleware_, where you do not want to make any assumptions about which PSR-7 implementation a consumer is using.

weierophinney:
Thanks, @gsteel!

PR #288, opened 2018-01-16
Do not include coverage uploading in composer.json

weierophinney:
Coverage reporting is only used on a single target, and is only installed conditionally via Travis, so we should not have scripts for uploading coverage within our `composer.json`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,7 @@ script:
   - if [[ $CHECK_CS == 'true' ]]; then composer cs-check ; fi
 
 after_script:
-  - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry composer upload-coverage ; fi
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -70,7 +70,7 @@ script:
   - if [[ $CHECK_CS == 'true' ]]; then composer cs-check ; fi
 
 after_script:
-  - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry composer upload-coverage ; fi
+  - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry php vendor/bin/php-coveralls -v ; fi
```
webimpress: Can't we have just:
` - if [[ $TEST_COVERAGE == 'true' ]]; then vendor/bin/php-coveralls -v ; fi`
as https://github.com/zendframework/maintainers/blob/master/template/.travis.yml#L52

webimpress: Ok, fine, I'll update then maintainers repo with these changes.

PR #290, opened 2018-02-05
check fstat result

pine3ree:
`fstat` calls may return `false` (remote resources, `'php://input'` streams)

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -103,7 +103,11 @@ class Stream implements StreamInterface
         }
 
         $stats = fstat($this->resource);
-        return $stats['size'];
+        if ($stats !== false) {
+            return $stats['size'];
+        }
+
+        return null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pine3ree:
I previously added a test case for remote resource (https://github.com/zendframework/zend-diactoros/pull/290/commits/7c44b5a2f29bb2bec5d96de202f1906594404a2a#diff-b0c96a6133152b0c2e2b5e59291b6c0a) that caused 2~3 connection timeouts in Travis, so I removed it.
I was thinking of keeping the test case handling the timeouts errors, but in that case we have no open pointer to pass to fstat, so that the test itself was pointless. Any hints?

weierophinney:
Honestly, the test written is fine; the point is to see if we have file information before attempting to return a member of it, which this tests.

weierophinney:
Thanks, @pine3ree.

PR #291, opened 2018-02-11
remove superfluous $scheme empty test

pine3ree:
`$scheme` is never empty as it was initialized to `'http'`


weierophinney:
Thanks, @pine3ree.

PR #293, opened 2018-02-26
`getHost` returns always lowercase hostname (PSR-7 spec)

### LINKED ISSUE #292: URI
PSR7 requires the `getHost` method return a lowercase string of the host name, or an empty string. Currently the host is being passed without transformation.

Diactoros - Uri.php
https://github.com/zendframework/zend-diactoros/blob/369cc8888e8e79f04c52c6dcd77a35c8f25da746/src/Uri.php#L437-L463

PSR7 - UriInterface.php
https://github.com/php-fig/http-message/blob/a6764a43faa657cef3e367a66785f2fc980bdff5/src/UriInterface.php#L79-L90

Ref zendframework/zend-stratigility#157

webimpress:
Fixes #292

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -303,7 +303,7 @@ class Uri implements UriInterface
         }
 
         $new = clone $this;
-        $new->host = $host;
+        $new->host = strtolower($host);
 
         return $new;
     }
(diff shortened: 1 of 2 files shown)


weierophinney:
Thanks, @webimpress

PR #295, opened 2018-03-18
[v2] Removed emitters and server - moved to zend-httphandlerrunner

webimpress:
We have duplicated functionality in `zend-httphandlerrunner`.

In this PR classes from zend-httphandlerrunner are used instead, and local classes are removed.

This is just quick change, to start discussion what we should do with this package. Maybe the package should be splitted to smaller one, or just `Server` class should be moved to another (new?) repository?


weierophinney:
I've created a release-2.0 branch for this, and updated the PR to base off of that.

Some notes:

- Definitely want to remove the emitters. The point of creating zend-httphandlerrunner was to extract those so that they may be used with _any_ PSR-7 implementation (which has been requested a number of times previously).
- I'd argue we can also remove the `Server` class from v2. While interesting, it belongs in a separate package, and zend-httphandlerrunner's `RequestHandlerRunner` class largely fulfills that purpose already.

PR #304, opened 2018-05-30
Support swoole_http_request

ezimuel:
This PR adds a `ServerRequestFactory::fromSwoole` static function to transform a [Swoole](https://www.swoole.co.uk/) request (`swoole_http_request`) into a PSR-7 request.

Moreover, it provides a `Zend\Diactoros\Response\SwooleEmitter` to convert a PSR-7 response to `swoole_http_response`.

The idea is to offer the possibility to use [zend-diactoros](https://github.com/zendframework/zend-diactoros) (and Expressive) with Swoole out-of-the-box.


src/ServerRequestFactory.php
```diff
@@ -114,10 +114,10 @@ public static function fromSwoole(swoole_http_request $request)
         if (! extension_loaded('swoole')) {
             throw new Exception\RuntimeException('Swoole extension is not installed!');
         }
-        $get    = $request->get ?? [];
```
Ocramius: AAAAAAAAAAAAAAAAA WE STILL SUPPORT PHP 5?!

src/ServerRequestFactory.php
```diff
@@ -99,6 +101,82 @@ public static function fromGlobals(
...
+                             $request->header['accept'] : '*/*',
+            'HTTP_ACCEPT_LANGUAGE' => isset($request->header['accept-language']) ?
+                                      $request->header['accept-language'] : '',
+            'HTTP_ACCEPT_ENCODING' => isset($request->header['accept-encoding']) ?
+                                      $request->header['accept-encoding'] : '',
+            'HTTP_CONNECTION' => isset($request->header['connection']) ?
+                                 $request->header['connection'] : '',
+            'HTTP_CACHE_CONTROL' => isset($request->header['cache-control']) ?
+                                    $request->header['cache-control'] : '',
+        ];
```
weierophinney: I'd argue we shouldn't populate most of the above under Swoole; most of these are things that an SAPI populates _because there is no other way to get at headers, the request URI parts, etc._. Since we _do_ have a way to get at those, I'd argue we should only populate the `SERVER_*` and `REMOTE_*` values, and maybe the `GATEWAY_INTERFACE` and `REQUEST_TIME`.

Also, please align these as follows:

```php
'key' => isset(...)
    ? ...
    : ...,
```

When it comes to marshaling the URI, we have the various parts in the Swoole request, and can pass a smaller array _in situ_ when calling `marshalUriFromServer()`:

```php
static::marshalUriFromServer([ /* values we need here */ ], $headers)
```

src/ServerRequestFactory.php
```diff
@@ -99,6 +101,82 @@ public static function fromGlobals(
...
+        foreach ($request->header as $name => $value) {
+            $headers[str_replace('-', '_', $name)] = $value;
+        }
+
+        return new ServerRequest(
+            $server,
+            static::normalizeFiles($files),
+            static::marshalUriFromServer($server, $headers),
+            $server['REQUEST_METHOD'],
+            $request->rawContent(),
```
weierophinney: Is there a possibility to retrieve a stream resource, by any chance? If so, we could decorate that in a `Stream` instance, which would be better.

src/ServerRequestFactory.php
ezimuel: This is the `var_dump` of a Swoole HTTP request, using this simple [test example](https://gist.github.com/ezimuel/c2613f771c415678dc50d1d1930585cd).

```
object(Swoole\Http\Request)#7 (9) {
  ["fd"]=>
  int(1)
  ["header"]=>
  array(5) {
    ["host"]=>
    string(14) "localhost:9501"
    ["user-agent"]=>
    string(12) "HTTPie/0.9.8"
    ["accept-encoding"]=>
    string(13) "gzip, deflate"
    ["accept"]=>
    string(3) "*/*"
    ["connection"]=>
    string(10) "keep-alive"
  }
  ["server"]=>
  array(11) {
    ["request_method"]=>
    string(3) "GET"
    ["request_uri"]=>
    string(1) "/"
    ["path_info"]=>
    string(1) "/"
    ["request_time"]=>
    int(1527775742)
    ["request_time_float"]=>
    float(1527775742.4968)
    ["server_port"]=>
    int(9501)
    ["remote_port"]=>
    int(32922)
    ["remote_addr"]=>
    string(9) "127.0.0.1"
    ["master_time"]=>
    int(1527775742)
    ["server_protocol"]=>
    string(8) "HTTP/1.1"
    ["server_software"]=>
    string(18) "swoole-http-server"
  }
  ["request"]=>
  NULL
  ["cookie"]=>
  NULL
  ["get"]=>
  NULL
  ["files"]=>
  NULL
  ["post"]=>
  NULL
  ["tmpfiles"]=>
  NULL
}
```
I think we should include all these values, what do you think?

src/ServerRequestFactory.php
ezimuel: It seems that `swoole_http_request` contains only the body message as string.

ezimuel:
@weierophinney I reduced the parameters for converting a Swoole request to PSR-7. I followed the `var_dump` output (see my previous comment) of a Swoole request.

ezimuel:
@weierophinney, @Ocramius I also added the Swoole emitter. We can now convert a `swoole_http_request` to PSR-7 and a PSR-7 response to `swoole_http_response`.

src/ServerRequestFactory.php
weierophinney: This definitely won't work as-is. `ServerRequestFactory::initialize()` accepts the following types of `$body` arguments:

- A `StreamInterface` instance.
- A string representing a stream URI to open.
- A PHP resource.

Passing just the contents will not work; it needs to be wrapped in a `Stream` instance first.

ezimuel:
@weierophinney Just added the documentation.

doc/book/swoole.md
```diff
@@ -0,0 +1,54 @@
+# Swoole
+use Zend\Diactoros\ServerRequestFactory;
```
weierophinney: I think this line belongs above line 15?

src/ServerRequestFactory.php
```diff
@@ -99,6 +101,52 @@ public static function fromGlobals(
...
+        foreach ($request->server as $key => $value) {
+            $server[strtoupper($key)] = $value;
+        }
+
+        $headers = [];
+        foreach ($request->header as $name => $value) {
+            $headers[str_replace('-', '_', $name)] = $value;
+        }
+
+        $body = new Stream('php://temp', 'wb+');
```
weierophinney: I've been thinking about this, and I'm worried this could be a performance issue and/or security vector in some cases; in particular, requests that include file uploads and/or requests with large requests bodies.

As such, I'm thinking we should create a proxy decorator class for the swoole request body that extends the current default `Stream` implementation:

- It would receive the request to the constructor.
- Any method that would normally access the underlying stream resource would:
  - check if the request body has been requested before
  - if not, it would create the `php://temp` stream and write the request body to it, and assign it via a call to `parent::construct()`.
  - otherwise, it would proxy to the parent method

In this way, we prevent duplication of the request contents _until they are accessed_, which alleviates the majority of issues (particularly since the file uploads will normally already be handled).

ezimuel:
@weierophinney I created a `SwooleStream` class implementing the `StreamInterface` with the `rawcontent()` proxy usage that you proposed. I also improved `ServerRequestFactory::fromSwoole()`  with additional checks.
Moreover, I removed the `SwooleEmitter` class in favor of a `zendframework/zend-expressive-swoole` component that I'll release soon.

doc/book/swoole.md
```diff
@@ -0,0 +1,16 @@
...
+
+[Swoole](https://www.swoole.co.uk/) is an async programming Framework for PHP
+that can be used to create high performance HTTP server applications, e.g. web
+APIs. We provided the support of Swoole in `zend-diactoros` using a static
+method to convert a [Swoole\Http\Request](http://php.net/manual/en/class.swoole-http-request.php)
+in a [PSR-7 Request](https://www.php-fig.org/psr/psr-7/#32-psrhttpmessagerequestinterface).
+
+This method is as follows:
+
+```
```
malukenho: <code>\`\`\`php</code>?

src/ServerRequestFactory.php
```diff
@@ -99,6 +101,52 @@ public static function fromGlobals(
...
     }
 
+    /**
+     * Create a request from a Swoole request
+     *
+     * @param swoole_http_request $request
+     * @return ServerRequest
+     * @throws RuntimeException if Swoole is not installed
+     */
+
```
malukenho: Remove empty line

src/SwooleStream.php
```diff
@@ -0,0 +1,189 @@
...
+ * @license   https://github.com/zendframework/zend-diactoros/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Diactoros;
+
+use Psr\Http\Message\StreamInterface;
+use RuntimeException;
+use swoole_http_request;
+
+class SwooleStream implements StreamInterface
```
malukenho: Why is it not final?

test/SwooleStreamTest.php
```diff
@@ -0,0 +1,179 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-diactoros for the canonical source repository
+ * @copyright Copyright (c) 2015-2017 Zend Technologies USA Inc. (http://www.zend.com)
```
malukenho: 2018?

test/SwooleStreamTest.php
```diff
@@ -0,0 +1,179 @@
...
+use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\StreamInterface;
+use swoole_http_request;
+use Zend\Diactoros\SwooleStream;
+
+class SwooleStreamTest extends TestCase
+{
+    const DEFAULT_CONTENT = 'This is a test!';
+
+    public function setUp()
```
malukenho: Shouldn't it be `protected` to respect the interface?

Moln:
Uh..... split to `zend-diactoros-swoole`, please.

src/ServerRequestFactory.php
```diff
@@ -99,6 +101,52 @@ public static function fromGlobals(
...
+    public static function fromSwoole(swoole_http_request $request)
+    {
+        if (! extension_loaded('swoole')) {
```
Ocramius: Doesn't look like a good idea: as commented by @moln, this should be a hard requirement, with an adapter package enforcing it

ezimuel: @Ocramius, @weierophinney maybe [zend-expressive-swoole](https://github.com/zendframework/zend-expressive-swoole/) is a better fit for this PR?

ezimuel:
@weierophinney, @Ocramius, @Moln  I already prepared #306 to move the Swoole outside `zend-diactoros` (e.g. in `zend-expressive-swoole`).

src/ServerRequestFactory.php
Ocramius: :+1:

ezimuel:
Closing this PR and moving it to [zend-expressive-swoole](https://github.com/zendframework/zend-expressive-swoole), waiting for #306 merge.

PR #306, opened 2018-06-26
Static functions to public in ServerRequestFactory

ezimuel:
This PR changes the private static functions in `ServerRequestFactory` to public, in order to be used by third-party libraries such as [zend-expressive-swoole](https://github.com/zendframework/zend-expressive-swoole). This PR will allows to move #304 in `zend-expressive-swoole` library.


weierophinney:
Per a conversation with @ezimuel , closing this in favor of #307.

PR #309, opened 2018-07-04
reduce $crFound check at AbstractSerializer::getLine()

samsonasik:


weierophinney:
I disagree with this change. The four distinct conditionals in the `while` loop are for a reason: to detail the specific carriage return/line feed combinations we scan for, and the impact each has when detecting a line. Moving one part of the conditional inside the other makes understanding the _combinations_ more difficult.

PR #310, opened 2018-07-07
Fixes #152 : SSL should keep :443 port for after host if it specified

### LINKED ISSUE #152: Host Request header misses port number for SSL connections
It seems the Host header calculated without a port number when the Request contains these values.

```
'REQUEST_SCHEME' => 'https',
...
'SERVER_PORT' => '443',
...
'HTTP_HOST' => 'some-domain.localhost',
...
'HTTPS' => 'on',
```

Host should be known as : `some-domain.localhost:443`

but is known as : `some-domain.localhost`

see https://github.com/neomerx/cors-psr7/issues/12

dump of $request https://gist.github.com/basz/0408e7e72d4458afb008

samsonasik:
Fixes #152


weierophinney:
Deferring for now, [per my comment on the original issue](https://github.com/zendframework/zend-diactoros/issues/152#issuecomment-403603844).

weierophinney:
Closing, per @basz.

Issue #311, opened 2018-07-08
Nested uploaded files parsed wrong in 1.8.0


aimeos:
There's a problem in 1.8.0 with nested file uploads for e.g.:
```
<input type="file" name="image[0][file]" />

```
Then, `createUploadedFile()` is called with that array:
```
Array (
	[image] => Array (
		[name] => Array ( [0] => Array ( [file] => admin-product-detail.png ) )
		[type] => Array ( [0] => Array ( [file] => image/png ) )
		[tmp_name] => Array ( [0] => Array ( [file] => /tmp/php9tIgbR ) )
		[error] => Array ( [0] => Array ( [file] => 0 ) )
		[size] => Array ( [0] => Array ( [file] => 83760 ) ) 
	) 
)
```
In version 1.7.2 nested files are correctly parsed

precariouspanther:
Noticed this too - diactoros is allowing single dimension arrays of files (i.e `image[0]`) and named indexes (i.e `image[my-image]`) but not multi-dimensional arrays (i.e `image[0][file]`, `image[slide][0][image][4]`) which are legal names for file upload elements.

precariouspanther:
Candidate fix PR here: https://github.com/zendframework/zend-diactoros/pull/312

PR #312, opened 2018-07-09
#311 - Update NormalizeUploadedFiles to handle multi-dimension recursively

precariouspanther:
To address https://github.com/zendframework/zend-diactoros/issues/311 (Some/All?) SAPI's index the `tmp_name` etc common fields from the point that a numerically indexed array is encountered which is causing Diactoros to fail fatally with form uploads that use multi-dimensional named fields containing a numeric index (as well as not converting all leaf nodes for deeper hierarchies of file upload field names).

This change updates normalizeUploadedFiles to be recursive and introduces several tests to verify the behaviour matches the PSR-7 spec + the scenario not mentioned by the spec.

Note: given the number of "private" functions being encapsulated in this larger function it may be worth considering converting this to an object instead especially as the recursion and lack of hoisting makes declaration order of the "private" functions non-idiomatic for most PHP engineers. I avoided significant changes here given it's just a fix, but food for thought!

Also, first time I've contributed to this project: please let me know if I've incorrectly interpreted the contrib. standards and I need to fix anything 😃 

@pascalzajac @serroba FYI

### DIFF
--- a/src/functions/normalize_uploaded_files.php
+++ b/src/functions/normalize_uploaded_files.php
@@ -25,6 +25,8 @@ use function is_array;
 function normalizeUploadedFiles(array $files)
 {
     /**
+     * Traverse a nested tree of uploaded file specifications.
+     *
      * @param string[]|array[] $tmpNameTree
      * @param int[]|array[] $sizeTree
      * @param int[]|array[] $errorTree
--- a/src/functions/normalize_uploaded_files.php
+++ b/src/functions/normalize_uploaded_files.php
@@ -25,6 +25,45 @@ use function is_array;
 function normalizeUploadedFiles(array $files)
 {
     /**
+     * @param string[]|array[] $tmpNameTree
+     * @param int[]|array[] $sizeTree
+     * @param int[]|array[] $errorTree
+     * @param string[]|array[]|null $nameTree
+     * @param string[]|array[]|null $typeTree
+     * @return UploadedFile[]|array[]
+     */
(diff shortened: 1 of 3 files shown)


src/functions/normalize_uploaded_files.php
```diff
@@ -24,6 +24,45 @@
  */
 function normalizeUploadedFiles(array $files)
 {
+    /**
+     * @param array $tmpNameTree
```
Ocramius: Could you please document the types of these arrays?

precariouspanther: Will do 👍 This was mimicking the existing function (as they are multi-dimensional) - would you like me to update the existing one too?

src/functions/normalize_uploaded_files.php
```diff
@@ -24,6 +24,45 @@
...
+    ) use (&$recursiveNormalize) {
+        $normalized = [];
+        foreach ($tmpNameTree as $key => $value) {
+            if (is_array($value)) {
+                // Traverse
+                $normalized[$key] = $recursiveNormalize(
+                    $tmpNameTree[$key],
+                    $sizeTree[$key],
+                    $errorTree[$key],
+                    $nameTree[$key] ?? null,
```
weierophinney: We're currently still supporting PHP 5.6, so this needs to become:

```php
isset($nameTree[$key]) ? $nameTree[$key] : null
```

(Here and everywhere else you use null coalesce.)

PR #313, opened 2018-07-09
Updates reason phrase for 425 response status code

weierophinney:
The IANA HTTP status codes list was updated on 2018-07-02 to reflect a new IETF draft, draft-ietf-httpbis-replay-04, which details a header that may be used to indicate that an "early data" request may not yet be accepted, as no TLS handshake has been previously made, and that the request myust be retried after receiving the 425 response.

Without this change, tests fail on all platforms as the hard-coded reason phrase in `Zend\Diactoros\Response` references an early, non-standards-track usage of the status code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Changed
 
-- Nothing.
+- [#313](https://github.com/zendframework/zend-diactoros/pull/313) changes the reason phrase associated with the status code 425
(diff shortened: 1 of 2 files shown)


samsonasik:
@weierophinney should `zend-http` Response class be updated as well ? https://github.com/zendframework/zend-http/blob/master/src/Response.php#L143

weierophinney:
@samsonasik Yes, please!

samsonasik:
@weierophinney I've created PR https://github.com/zendframework/zend-http/pull/153 for it

PR #314, opened 2018-07-09
Workaround PHP bug #63206 in Stream::setStream method

### LINKED ISSUE #298: Stream::setStream method causes PHP bug #63206
While using Zend Diactoros from inside an error handler I found out that the next errors were not logged anymore through my custom error handler. The cause is the PHP bug [#63206](https://bugs.php.net/bug.php?id=63206) raised by the following code:

https://github.com/zendframework/zend-diactoros/blob/bb833f8e92297be103bf6cd3fc25219474bbf0d0/src/Stream.php#L310-L314

There are two workarounds that I can think of: do not use the second argument of the `set_error_handler` method or suppress the errors reported by the `fopen` method and check instead if it returns `false`

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

This PR fixes #298 by removing the usage of the second argument of the `set_error_handler` function and moves the check of the level of the raised error inside the error handler callback.

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -330,8 +330,12 @@ class Stream implements StreamInterface
 
         if (is_string($stream)) {
             set_error_handler(function ($e) use (&$error) {
+                if ($e !== E_WARNING) {
+                    return;
+                }
+
                 $error = $e;
-            }, E_WARNING);
+            });
             $resource = fopen($stream, $mode);
             restore_error_handler();
         }


weierophinney:
@ste93cry Umm... that's not the PR template we use... and the package is licensed BSD-3-Clause; we're not in a position to dual-license. Can you clarify the license before we merge, please?

ste93cry:
Sorry I copy-pasted a template from another repository I used to contribute to. My code is released under any license you use, so feel free to do whatever you want.

weierophinney:
Thanks, @ste93cry!

PR #319, opened 2018-07-19
Status code must be an integer following psr-7

snapshotpl:


src/Response.php
```diff
@@ -174,8 +174,7 @@ public function withStatus($code, $reasonPhrase = '')
      */
     private function setStatusCode($code)
     {
-        if (! is_numeric($code)
-            || is_float($code)
+        if (! is_int($code)
```
weierophinney: One potential problem: the status code value may be derived from a string. As an example, it may come from the response line itself:

```http
HTTP/1.1 200 OK
```

In that case, we previously allowed doing something like the following:

```php
[$protocolAndVersion, $statusCode, $reasonPhrase] = explode(' ', $statusLine, 3);
```

and passing the values directly to the constructor and/or `withStatus()`.

With this change, the consumer is now forced to cast to `int` _before_ doing so, which becomes a BC break.

src/Response/Serializer.php
```diff
@@ -106,6 +106,6 @@ private static function getStatusLine(StreamInterface $stream)
             throw new UnexpectedValueException('No status line detected');
         }
 
-        return [$matches['version'], $matches['status'], isset($matches['reason']) ? $matches['reason'] : ''];
+        return [$matches['version'], (int) $matches['status'], isset($matches['reason']) ? $matches['reason'] : ''];
```
weierophinney: This is exactly the point I was making previously; with the change you suggest above, we now have a BC break requiring consumers make changes such as this.

test/ResponseTest.php
```diff
@@ -78,7 +78,7 @@ public function ianaCodesReasonPhrasesProvider()
         $records = $xpath->query('//ns:record');
 
         foreach ($records as $record) {
-            $value = $xpath->query('.//ns:value', $record)->item(0)->nodeValue;
+            $value = (int) $xpath->query('.//ns:value', $record)->item(0)->nodeValue;
```
weierophinney: This also confirms the BC break; previously, casting to (int) was unnecessary.

snapshotpl:
@weierophinney can we keep this changes to v2.0?

snapshotpl:
That's explanation is enough for me. I will prepare changes

weierophinney:
Thanks, @snapshotpl!

PR #320, opened 2018-07-19
Reason phrase must be a string

snapshotpl:


src/Response.php
```diff
@@ -186,6 +180,19 @@ private function setStatusCode($code)
                 static::MAX_STATUS_CODE_VALUE
             ));
         }
+
+        if (! is_string($reasonPhrase)) {
+            throw new InvalidArgumentException(sprintf(
+                'Unsupported HTTP method; must be a string, received %s',
```
weierophinney: "Unsupported HTTP method"?

I'm thinking you mean "Unsupported response reason phrase" here.

src/Response.php
```diff
@@ -186,6 +180,19 @@ private function setStatusCode($code)
...
+                (is_object($reasonPhrase) ? get_class($reasonPhrase) : gettype($reasonPhrase))
```
weierophinney: You can drop the surrounding parens; they're unneeded when providing an expression as a function argument.

test/ResponseTest.php
```diff
@@ -112,6 +112,13 @@ public function testCanSetCustomReasonPhrase()
         $this->assertSame('Foo Bar!', $response->getReasonPhrase());
     }
 
+    public function testCannotCreateReasonPhraseNotString()
```
weierophinney: Let's rephrase this to `testWithStatusRaisesAnExceptionForNonStringReasonPhrases()`. Additionally, please test it with other non-string values as well (using a data provider), to validate they all produce an exception.

weierophinney:
Thanks, @snapshotpl!

PR #322, opened 2018-07-19
Support for phpunit 7

snapshotpl:


test/PhpInputStreamTest.php
```diff
@@ -36,10 +36,10 @@ public function getFileContents()
         return file_get_contents($this->file);
     }
 
-    public function assertStreamContents($test, $message = null)
+    public function assertStreamContents($test)
     {
         $content = $this->getFileContents();
-        $this->assertSame($content, $test, $message);
+        $this->assertSame($content, $test);
     }
```
weierophinney: Why were the changes above made? `assertSame()` still allows providing a third `$message` argument...

snapshotpl: Yes, but in phpunit 7 is typed as string and null is not allowed anymore

weierophinney:
Thanks, @snapshotpl!

PR #323, opened 2018-07-19
Improvement withUri

snapshotpl:
Avoid heavy string operations with `strtolower`

### DIFF
--- a/src/RequestTrait.php
+++ b/src/RequestTrait.php
@@ -251,10 +251,13 @@ trait RequestTrait
      */
     public function withUri(UriInterface $uri, $preserveHost = false)
     {
+        $normalizedHeader = 'host';
         $new = clone $this;
         $new->uri = $uri;
 
-        if ($preserveHost && $this->hasHeader('Host')) {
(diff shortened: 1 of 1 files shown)


src/MessageTrait.php
```diff
@@ -262,14 +274,18 @@ public function withAddedHeader($header, $value)
      */
     public function withoutHeader($header)
     {
-        if (! $this->hasHeader($header)) {
-            return clone $this;
+        $normalized = strtolower($header);
+
+        if ($normalized === 'host') {
+            return $this;
         }
```
weierophinney: This change doesn't make sense to me. The `Host` header is specific to requests, which is why that class had additional logic present for handling it. Can you explain the rationale you have in making this change, as well as the changes to `Request` and `RequestTrait`, please? I'm not following what the changes accomplish, and it looks like they could potentially affect `Response` messages as well as currently proposed.

weierophinney:
@snapshotpl Ping... Please see my review comment, and either address the feedback or close.

Thanks!

weierophinney:
Closing due to inactivity.

PR #325, opened 2018-08-12
Add integration tests

Nyholm:
I noticed that this library does not fully comply with psr7. 

I’ve added integration test to show where it is failing. 

(I’m at the airport atm, will complete this PR later)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ cache:
 env:
   global:
     - COMPOSER_ARGS="--no-interaction"
-    - COVERAGE_DEPS="php-coveralls/php-coveralls"
+    - DEPENDENCIES=""
     - LEGACY_DEPS="phpunit/phpunit"
 
 matrix:
@@ -31,6 +31,7 @@ matrix:
         - DEPS=locked
         - CHECK_CS=true
         - TEST_COVERAGE=true
+        - DEPENDENCIES="php-coveralls/php-coveralls"
     - php: 7
       env:
         - DEPS=latest
@@ -52,6 +53,11 @@ matrix:
     - php: 7.2
       env:
         - DEPS=latest
+    - php: 7.2
+      name: Integration tests
+      env:
+        - DEPENDENCIES="http-interop/http-factory-diactoros"
+
 
 before_install:
   - if [[ $TEST_COVERAGE != 'true' && "$(php --version | grep xdebug -ci)" -ge 1 ]]; then phpenv config-rm xdebug.ini || return 0 ; fi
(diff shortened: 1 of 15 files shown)


Nyholm:
Oh. It looks like this PR is working as expected. There are just some cs fixes needed

Ocramius:
Picking this up for checking while I'm on travel tomorrow :+1:

.travis.yml
```diff
@@ -9,7 +9,7 @@ cache:
 env:
   global:
     - COMPOSER_ARGS="--no-interaction"
-    - COVERAGE_DEPS="php-coveralls/php-coveralls"
+    - DEPENDENCIES=""
```
weierophinney: Technically, we don't need to pre-define globals; the shell will resolve it to `''` if not defined.

.travis.yml
```diff
@@ -62,7 +68,7 @@ install:
   - if [[ $TRAVIS_PHP_VERSION =~ ^5.6 ]]; then travis_retry composer update $COMPOSER_ARGS --with-dependencies $LEGACY_DEPS ; fi
   - if [[ $DEPS == 'latest' ]]; then travis_retry composer update $COMPOSER_ARGS ; fi
   - if [[ $DEPS == 'lowest' ]]; then travis_retry composer update --prefer-lowest --prefer-stable $COMPOSER_ARGS ; fi
-  - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry composer require --dev $COMPOSER_ARGS $COVERAGE_DEPS ; fi
+  - if ! [ -v "$DEPENDENCIES" ]; then travis_retry composer require --dev $COMPOSER_ARGS $DEPENDENCIES ; fi
```
weierophinney: This should become `if [[ $DEPENDENCIES != '' ]]` to be consistent with our other checks.

Also, please rename it to `INTEGRATION_DEPS` so we know _what kind_ of dependencies we're defining and using here.

test/Integration/RequestTest.php
```diff
@@ -0,0 +1,28 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-diactoros for the canonical source repository
+ * @copyright Copyright (c) 2015-2017 Zend Technologies USA Inc. (http://www.zend.com)
```
weierophinney: Change this line in all new files to:

```php
 * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (https://www.zend.com)
```

(Copyright extends from when the _file_ is introduced. Also, updates to use SSL-enabled URL.)

PR #330, opened 2018-09-08
Removed duplicated strict type declaration

webimpress:
There was duplicated strict type declaration in that file.


weierophinney:
Thanks, @webimpress!

PR #337, opened 2018-11-24
Fixed serverRequest factory

oscarotero:
This fixes the `ServerRequestFactory` that always creates a server request with `php://input` as the body stream.

### DIFF
--- a/test/ServerRequestFactoryTest.php
+++ b/test/ServerRequestFactoryTest.php
@@ -570,4 +570,15 @@ class ServerRequestFactoryTest extends TestCase
         $uri = marshalUriFromSapi($server, $headers);
         $this->assertSame('/requested/path', $uri->getPath());
     }
+
+    public function testServerRequestFactoryHasAWritableEmptyBody()
+    {
+        $factory = new ServerRequestFactory();
+        $request = $factory->createServerRequest('GET', '/');
+        $body = $request->getBody();
+
+        $this->assertTrue($body->isWritable());
+        $this->assertTrue($body->isSeekable());
+        $this->assertSame(0, $body->getSize());
+    }
 }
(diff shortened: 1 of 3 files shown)


Ocramius:
This requires a unit test, as well as an integration test that runs through a standard HTTP request flowing in from at least the internal PHP development server.

oscarotero:
Ok, I've added a couple of tests.

test/ServerRequestFactoryTest.php
```diff
@@ -570,4 +570,15 @@ public function testMarshalRequestUriPrefersRequestUriServerParamWhenXOriginalUr
...
         $this->assertSame('/requested/path', $uri->getPath());
     }
+
+    public function testServerRequestFactoryHasAWritableEmptyBody()
+    {
+        $factory = new ServerRequestFactory();
+        $request = $factory->createServerRequest('GET', '/');
+        $body = $request->getBody();
+
+        $this->assertTrue($body->isWritable());
```
Ocramius: Hmm, I don't think the global request body should be writable here

oscarotero: That's not the "global request body" but a body of a server request just created without global values. To me, creating an empty `php://temp` body that cannot be writable makes no sense.

Ocramius: This is the request coming in from global state: if anybody is trying to write to this body, something went extremely wrong IMO.

oscarotero: What?! `createServerRequest` should create a server request using only the values passed in the constructor (method, uri and serverParams), and never use global variables. There's the `fromGlobals` to do that. Or I'm missing something?

An example of use case [is testing](https://github.com/middlewares/payload/blob/master/tests/PayloadTest.php#L33).

Ocramius: Ah yes, I got confused here, my mistake!

PR #339, opened 2018-12-04
Updated export-ignores to cleanup release

JoeBengalen:
Not sure how this fits the PR template, so I left it out.

I noticed some files in the downloaded package from packagist that should not be there.

image

Result should be
```
src/
CHANGELOG.md
LICENSE.md
README.md
composer.json
```

Proposed contents are based on other zend projects like
https://github.com/zendframework/zend-httphandlerrunner/blob/master/.gitattributes
https://github.com/zendframework/zend-expressive/blob/master/.gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,9 +1,12 @@
-/doc export-ignore
+/docs export-ignore
 /test export-ignore
 /vendor export-ignore
+.coveralls.yml export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore
 .scrutinizer.yml export-ignore
 .travis.yml export-ignore
+composer.lock export-ignore
+mkdocs.yml export-ignore
 phpunit.xml.dist export-ignore
 phpcs.xml export-ignore


.gitattributes
```diff
@@ -1,9 +1,11 @@
-/doc export-ignore
-/test export-ignore
-/vendor export-ignore
-.gitattributes export-ignore
-.gitignore export-ignore
-.scrutinizer.yml export-ignore
-.travis.yml export-ignore
-phpunit.xml.dist export-ignore
-phpcs.xml export-ignore
+/.coveralls.yml export-ignore
```
Ocramius: Could you please re-do this patch without changing the sorting of entries? This diff is really noisy, plus `git blame` information is lost.

JoeBengalen: Can do. I put it in the same order as the linked projects have.
Would you also like it without the `/` prefix?

Ocramius: Simply leave the ones that weren't changed as they are. If the `/` is not present in existing ones, we should also not add it here

JoeBengalen:
All done, thnx for the quick answers :)

Ocramius:
@JoeBengalen ah, I just realised one last thing, sorry - can you please change the target to `develop`, and rebase on that? This is an improvement, and all improvements go to `develop`.

JoeBengalen:
Took a google search, but I think I got it haha

Issue #342, opened 2018-12-15
ServerRequest: Header with value "0" discarded


TCB13:
- [x] I was not able to find an [open](https://github.com/zendframework/zend-diactoros/issues?q=is%3Aopen) or [closed](https://github.com/zendframework/zend-diactoros/issues?q=is%3Aclosed) issue matching what I'm seeing.
 - [x] This is not a question. (Questions should be asked on [chat](https://zendframework.slack.com/) ([Signup here](https://zendframework-slack.herokuapp.com/)) or our [forums](https://discourse.zendframework.com/c/questions/expressive).)

When a request is parsed with `ServerRequestFactory` it discards any header set as `0`. 

### Code to reproduce the issue

Create a `client.php`:
```php
$ch = curl_init();
curl_setopt($ch, CURLOPT_URL, $url);
curl_setopt($ch, CURLOPT_CUSTOMREQUEST, "PATCH");
curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
curl_setopt($ch, CURLOPT_HEADER, true);
curl_setopt($ch, CURLOPT_HTTPHEADER, ["Upload-Offset: 0"]);
curl_exec($ch);
```
Create a `server.php` and ask for the `Upload-Offset` header:
```php
$request = Zend\Diactoros\ServerRequestFactory::fromGlobals();
$result = $request->getHeaderLine("Upload-Offset");
```
### Expected results
`$result` should be `0`.

### Actual results
`$result` is set to empty string.

Testing this via xDebug to compare the result of the built in `getallheaders()` with `->getHeaderLine` shows that the issue is in this package:
![headers](https://user-images.githubusercontent.com/6315832/50045252-5ea57f00-0087-11e9-94ed-276d80acc6a2.png)

This issue actually makes it impossible to implement a protocol that requires `Upload-Offset` to always be set even if to zero.

Maybe somehow related to https://github.com/zendframework/zend-http/issues/26 ?

Thank you.

malukenho:
Hello @TCB13!
I've been investigating this issue and I found that whatever header you pass with the value `0` is being removed. You can try with the header `accept` for example.

Without passing the `0` as the header value:
```sh
$ curl -H http://localhost:8089/server.php

# Array
# (
#     [host] => localhost:8089
#     [user-agent] => curl/7.54.0
#     [accept] => */*
# )
```

```sh
$ curl -H 'Accept: 0' http://localhost:8089/server.php

# Array
# (
[...]

PR #344, opened 2018-12-19
Bugfix: Header with value "0" is being discarded

malukenho:
- [x] Are you fixing a bug?

Linking #342

### DIFF
--- a/src/functions/marshal_headers_from_sapi.php
+++ b/src/functions/marshal_headers_from_sapi.php
@@ -46,6 +46,8 @@ function marshalHeadersFromSapi(array $server) : array
             $headers[$name] = $value;
             continue;
         }
+
+        $headers[$key] = $value;
     }
 
     return $headers;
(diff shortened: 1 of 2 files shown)


weierophinney:
Thanks, @malukenho!

tptrixtop:
This fix cause an issue in apache on win, because there is a header

[SERVER_SIGNATURE] =>`<address>Apache/2.4.34 (Win32) OpenSSL/1.1.0i PHP/7.2.9 Server at api.best-office.local Port 8080</address>`

That contains not "secured" data and it throws an exception and drop down our prod environment. It was really hard to find the root cause.

vakata:
This is not a proper fix - all server vars are copied as headers, which is incorrect. The problem users are reporting is when some of the server values can not be used as headers. This should probably be reverted and a simple ```strlen($val)``` check added in its place.

froschdesign:
@vakata 
Please look #349 and add your comment there. Thanks!

Issue #346, opened 2018-12-20
Version 2.1.0 breaks on Apache


Y0lk:
- [x] I was not able to find an [open](https://github.com/zendframework/zend-diactoros/issues?q=is%3Aopen) or [closed](https://github.com/zendframework/zend-diactoros/issues?q=is%3Aclosed) issue matching what I'm seeing.
 - [x] This is not a question. (Questions should be asked on [chat](https://zendframework.slack.com/) ([Signup here](https://zendframework-slack.herokuapp.com/)) or our [forums](https://discourse.zendframework.com/c/questions/expressive).)

Since the update to 2.1.0, I'm getting the following error on a local Apache install on Windows. I haven't tested this anywhere else yet.

> Fatal error: Uncaught Zend\Diactoros\Exception\InvalidArgumentException: "<address>Apache/2.4.9 (Win32) PHP/7.1.24 Server at artofwhere.local Port 80</address> " is not valid header value

The exact change that causes this issue is this commit: https://github.com/zendframework/zend-diactoros/commit/1e8b6c4db2aed9647945f86a3e46178fca509e04#diff-9d47206c200ee53ffe55d05493ddd000

I've tried commenting out the line that was added and it fixes the issue

### Code to reproduce the issue

```php
$request = \Zend\Diactoros\ServerRequestFactory::fromGlobals(
    $_SERVER,
    $_GET,
    $_POST,
    $_COOKIE,
    $_FILES
);
```

bluebaroncanada:
@Y0lk, you can fix this for now by deleting line 50 from `marshal_headers_from_sapi.php`.

Y0lk:
Yeah, as I wrote in the description, it's the commit that added this line 50 that causes problem. You can just force the version to 2.0.1 in your composer.json for a cleaner fix for now

bluebaroncanada:
Ah.  Sorry.  I actually came upon this issue after I made the pull request. Sorry I didn't read it well enough.  He was trying to fix an error where he was assuming headers with a 0 value were not being accepted.  He ended up defeating the filter in the process. I think he just needs to prefix his headers with `HTTP_`.

bluebaroncanada:
#349

PR #347, opened 2018-12-21
500 error with Apache

bluebaroncanada:

- [x] Are you fixing a bug?
Yes
Using Apache 2.4.29 making any request
Throws a 500 error: <address>Apache/2.4.29 (Ubuntu) Server at dev2.studentfleamarket.com Port 443</address> is not a valid header
Removes that header.
Okay.
Test already exists/not necessary
Done

- [x] Is this related to quality assurance?
Probably not.  Related to #346.
More information in the blame for #344

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@ All notable changes to this project will be documented in this file, in reverse 
 ### Fixed
 
 - [#344](https://github.com/zendframework/zend-diactoros/pull/344) provides a fix to ensure that headers with a value of "0" are retained.
-- Reversed fix incorrect fix for [#344]
+- Reversed incorrect fix for [#344]
 
 ## 2.0.1 - 2018-12-03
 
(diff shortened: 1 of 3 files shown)


bluebaroncanada:
More on this issue is in the blame for #344

Moln:
I have the same issue.

malukenho:
I'm not sure if we should revert it or tag a new version from the old hash meanwhile we introduce a proper fix for it.

bluebaroncanada:
I don't think it needs a fix.

tptrixtop:
This fix cause an issue in apache on win, because there is a header
`
    [SERVER_SIGNATURE] =>
<address>Apache/2.4.34 (Win32) OpenSSL/1.1.0i PHP/7.2.9 Server at api.best-office.local Port 8080</address>
`
That contains not "secured" data and it throws an exception and drop down our prod environment. It was really hard to find the root cause.

PR #349, opened 2018-12-21
Headers with value 0 being stripped out. Previous bugfix caused all headers to be accepted.

bluebaroncanada:

- [x] Are you fixing a bug?
```
$ch = curl_init();
curl_setopt($ch, CURLOPT_URL, 'http://xxx.xxx.xxx');
curl_setopt($ch, CURLOPT_CUSTOMREQUEST, "PATCH");
curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
curl_setopt($ch, CURLOPT_HEADER, true);
curl_setopt($ch, CURLOPT_HTTPHEADER, ["Upload-Offset: 0"]);
curl_exec($ch);
```
The header `Upload-Offset` is being stripped. (Any header with value "0")
The value was being used in a conditional.  To test `!!("0") === false`.
Test was incorrect because the header was being stripped because it was not prefixed with `HTTP_`.
Relates to #347 #342 #344 
Values with a string of "0" should be accepted
Test passes with proper header formation.

### DIFF
--- a/src/functions/marshal_headers_from_sapi.php
+++ b/src/functions/marshal_headers_from_sapi.php
@@ -35,19 +35,21 @@ function marshalHeadersFromSapi(array $server) : array
             }
         }
 
-        if ($value && strpos($key, 'HTTP_') === 0) {
+        if ($value === '') {
+            continue;
+        }
+
+        if (strpos($key, 'HTTP_') === 0) {
             $name = strtr(strtolower(substr($key, 5)), '_', '-');
             $headers[$name] = $value;
             continue;
         }
 
-        if ($value && strpos($key, 'CONTENT_') === 0) {
+        if (strpos($key, 'CONTENT_') === 0) {
             $name = 'content-' . strtolower(substr($key, 8));
             $headers[$name] = $value;
             continue;
         }
-
-        $headers[$key] = $value;
     }
 
     return $headers;
--- a/test/ServerRequestFactoryTest.php
+++ b/test/ServerRequestFactoryTest.php
@@ -60,6 +60,23 @@ class ServerRequestFactoryTest extends TestCase
         $this->assertSame($expected, marshalHeadersFromSapi($server));
     }
 
+    public function testMarshalInvalidHeadersStrippedFromServerArray()
+    {
+        $server = [
+            'COOKIE' => 'COOKIE',
+            'HTTP_AUTHORIZATION' => 'token',
+            'MD5' => 'CONTENT-MD5',
+            'CONTENT_LENGTH' => 'UNSPECIFIED',
+        ];
+
+        //Headers that don't begin with HTTP_ or CONTENT_ will not be returned
+        $expected = [
+            'authorization' => 'token',
+            'content-length' => 'UNSPECIFIED',
+        ];
+        $this->assertSame($expected, marshalHeadersFromSapi($server));
+    }
+
     public function testMarshalsVariablesPrefixedByApacheFromServerArray()
     {
         // Non-prefixed versions will be preferred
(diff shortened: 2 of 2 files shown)


test/ServerRequestFactoryTest.php
```diff
@@ -427,10 +427,10 @@ public function testFromGlobalsUsesCookieHeaderInsteadOfCookieSuperGlobal()
      */
     public function testCreateFromGlobalsShouldPreserveKeysWhenCreatedWithAZeroValue()
     {
-        $_SERVER['Accept'] = '0';
+        $_SERVER['HTTP_ACCEPT'] = '0';
 
         $request = ServerRequestFactory::fromGlobals();
-        $this->assertSame('0', $request->getHeaderLine('Accept'));
+        $this->assertSame('0', $request->getHeaderLine('ACCEPT'));
```
Ocramius: Please add a test scenario, instead of replacing an existing one.

bluebaroncanada: The old test is not valid, because the bugfix was spurious, but the function name takes the name that needs to be used.   This could be a revert and then I could resubmit, but I don't think that's really necessary; do you?

bluebaroncanada:
This is the if statement that filters out headers that don't begin with `HTTP_` in `marshal_headers_from_sapi.php` on line 38:
```
```
The problem with that is that if `$value === "0"` the statement will return false, which is the original reported issue.  You can test the truth of this by observing `!!("0")===false`.
My fix is to change that line to
```
        if (($value || $value === "0") && strpos($key, 'HTTP_') === 0) {
```
which accounts  for the possibility that the value could be "0", which is still a valid value.  It actually might be better to say `!empty($value)`  but I don't know what other side effects that might have.

I then changed your test header in from `Accept` to `HTTP_ACCEPT` and the assertion header to `ACCEPT`.  (The `HTTP_` gets stripped away.)

Maybe it is better to say `!empty($value)` because maaaaaaaaaaaybe you want to keep `"false"`.  Then again maybe not?

It could break anything that expected `"false"` to not return a header.  I doubt that's the behaviour we would want, though.  It could be possible other people are negatively affected by this.  For right now there's no issue posted with `"false"`, so I think this commit should stand.

webimpress:
@bluebaroncanada I think it would be good to add tests for all cases you described in your comment. It could prevent the issue in the future.

bluebaroncanada:
@webimpress Then I have to decide and I hesitate to be the one to decide because I'm not the original author and I don't know his or her intention.
Calling @weierophinney

bluebaroncanada:
@weierophinney Should headers with `"false"` or `"0"` be kept.  The original behaviour is to discard because we're checking the truth of `$value` which is a string.
Observe that `!!("0") === !!("false") === false`.

More information: http://php.net/manual/en/function.boolval.php

bluebaroncanada:
Well colour me stupid. `!!("false")` is true.  I thought I had read it was false.  Reading less stupidly, I realize it's not listed in that document and I ran it through a tester and it came out `true`.
Still, your attention to this bug might be helpful.

src/functions/marshal_headers_from_sapi.php
```diff
@@ -35,7 +35,7 @@ function marshalHeadersFromSapi(array $server) : array
             }
         }
 
-        if ($value && strpos($key, 'HTTP_') === 0) {
+        if (($value || $value === "0") && strpos($key, 'HTTP_') === 0) {
```
webimpress: What about `$value !== ''` here?

webimpress:
@bluebaroncanada As you are submitting the PR you have to decide how it should form in your opinion. And if you demonstrate it by tests others can see if it makes sense or not. I would change the condition to be `$value !== ''` and then add tests to cover it. So `0` and `false` will be kept and we should have tested these cases. Also we would need test with empty value.

src/functions/marshal_headers_from_sapi.php
webimpress: I think the same we should have in line 44, or maybe before:

```php
if ($value === '') {
    continue;
}
```

bluebaroncanada:
- Changed checking to `$value !== ''`
- Any talk about `'false'` in this thread can be ignored
- Added new tests to explicitly test that server variables not prefixed with `HEADER_` or `CONTENT_` are stripped
- Also applied the new checking method  to `CONTENT_` prefixed variables
- Added a second assertion to check that `CONTENT_` variables with value of `'0'` are kept

src/functions/marshal_headers_from_sapi.php
bluebaroncanada: I agree. Updated.

xtreamwayz:
I have just tested this locally and it fixes the issues mentioned in #350.

xtreamwayz:
Until this is merged you can use this to skip affected versions: `"zendframework/zend-diactoros": "^2.0 !=2.0.2 !=2.1.0",`

weierophinney:
This looks good. I'm going to squash the commits so that I can more easily cherry-pick the changes and create a 2.0.3 release; I'll push those back to your branch before merging.

bluebaroncanada:
How do you do that exactly? Would like to know.

On Sat, 5 Jan 2019 at 14:50, weierophinney <notifications@github.com<mailto:notifications@github.com>> wrote:

This looks good. I'm going to squash the commits so that I can more easily cherry-pick the changes and create a 2.0.3 release; I'll push those back to your branch before merging.

bluebaroncanada:
image

PR #352, opened 2019-01-03
fix (Stream) Fix the fatal error on parameter 2

farpat:

- [ x ] Are you fixing a bug?

When $length variable of Stream@read method is strictly a string (even if this one contains a number).

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -274,7 +274,7 @@ class Stream implements StreamInterface
             throw Exception\UnreadableStreamException::dueToConfiguration();
         }
 
-        $result = fread($this->resource, $length);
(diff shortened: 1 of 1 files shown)


Ocramius:
The API strictly declares the parameter to be an `int`, so do pass an integer to it.

Closing as `invalid`.

Please also consider installing a static analysis tool such as phpstan or psalm in your project, to prevent these issues upfront.

farpat:
Thanks for response. As you talk, I must a PR on an other repository.

Thanks :)

farpat:
For information, this fix is passed here : https://github.com/slimphp/Slim/pull/2558

PR #353, opened 2019-01-04
Fixed bug where all server vars are copied as headers

vakata:

- [x] Are you fixing a bug?
  - [x] Detail how the bug is invoked currently: when creating a new request from globals
  - [x] Detail the original, incorrect behavior: all server vars are copied as headers
  - [x] Detail the new, expected behavior: only actual headers are extracted from $_SERVER
  - [ ] Base your feature on the `master` branch, and submit against that branch.
  - [ ] Add a regression test that demonstrates the bug, and proves the fix.
  - [ ] Add a `CHANGELOG.md` entry for the fix.

### DIFF
--- a/src/functions/marshal_headers_from_sapi.php
+++ b/src/functions/marshal_headers_from_sapi.php
@@ -35,19 +35,17 @@ function marshalHeadersFromSapi(array $server) : array
             }
         }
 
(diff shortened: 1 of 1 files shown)


vakata:
Did not see #349 - sorry.

PR #355, opened 2019-02-14
PHPDBG is also a valid non-sapi environment

dakota:
When calling `moveTo()` on `UploadedFile` in a `phpdbg` environment, `move_uplaoded_file()` is incorrectly called (and as expected fails). This adds `phpdbg` as a valid "non-SAPI" environment (Same as `cli`)


weierophinney:
Thanks, @dakota ! Merged and released with 2.1.2.

Issue #357, opened 2019-04-03
Uncaught TypeError: Return value of Zend\\Diactoros\\Uri::filterQueryOrFragment() must be of the type string, null returned


YellowPepper:
```
PHP Fatal error:  Uncaught TypeError: Return value of Zend\\Diactoros\\Uri::filterQueryOrFragment() must be of the type string, null returned in /vendor/zendframework/zend-diactoros/src/Uri.php:660\n
Stack trace:\n
#0 /vendor/zendframework/zend-diactoros/src/Uri.php(618): Zend\\Diactoros\\Uri->filterQueryOrFragment('\\xCF')\n
#1 /vendor/zendframework/zend-diactoros/src/Uri.php(410): Zend\\Diactoros\\Uri->filterQuery('Keyword=\\xCF')\n
#2 /vendor/zendframework/zend-diactoros/src/functions/marshal_uri_from_sapi.php(218): Zend\\Diactoros\\Uri->withQuery('Keyword=\\xCF3')\n
```

Version
 zend-diactoros 2.1.1
 php 7.2.5
### Expected results

TypeError does not occur

krowinski:
+1

krowinski:
+1 

it's returning null because of some regexp error can we use preg_last_error to check and throw exception in filterQueryOrFragment ?

weierophinney:
Can you provide example `$server` and/or `$header` arrays that demonstrate the issue, please? From there we can write a test, which will help us create a fix and prevent a future regression.

Thanks!

krowinski:
@weierophinney Hi, here you go 

```
<?php
var_dump(
    preg_replace_callback(
        '/(?:[^a-zA-Z0-9_\-\.~\pL!\$&\'\(\)\*\+,;=%:@\/\?]+|%(?![A-Fa-f0-9]{2}))/u',
        static function (array $matches) {return rawurlencode($matches[0]);} ,
        ("\x21\x92")
    ), 
    preg_last_error()
);
```

http://sandbox.onlinephpfunctions.com/code/9d90fde0fac05f1d26c0f3310fee3e6cf2405b91

weierophinney:
@krowinski I'm not asking for a patch, I'm asking for a _reproduce case_. I need that so I can write a unit test...

krowinski:
what? this is not a patch its a test how to break preg match to return null, I will fork repo and write test myself

krowinski:
I added too tests example that will fail preg match here 

[...]

Issue #218, opened 2016-12-26
Character encoding issue in Uri.php filterQuery() method.


oroszlanyzsolt:
In example when the query string contains something like this: "?name=value&név=érték" in internet explorer the browser does not send it as utf-8 encoded, the result is "?name=value&n?v=?rt?k".
It is a known bug in [Internet Explorer browsers](https://blogs.msdn.microsoft.com/ieinternals/2014/04/22/unicode-in-url-changes-for-ie11/).
Possible workaround is checking return value of `filterQueryOrFragment()` for null value (because `filterQueryOrFragment()` returns null on failure) and call `filterQueryOrFragment()` again but the input have to run through `utf8_encode()`  function. All together: `filterQueryOrFragment(utf8_encode(<query_string_here>))`.
As far as I know other browsers are not affected by this bug, e.g.: Mozilla Firefox, Google Chrome, Opera browser (current versions).

PR #359, opened 2019-04-05
Bug fix: Uncaught TypeError: Return value of Zend\\Diactoros\\Uri::fi…

soru1001:
…lterQueryOrFragment() must be of the type string, null returned

- [x] Are you fixing a bug?
　#218 
　#357

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -652,7 +652,7 @@ class Uri implements UriInterface
     /**
      * Filter a query string key or value, or a fragment.
      */
-    private function filterQueryOrFragment(string $value) : string
+    private function filterQueryOrFragment(string $value) : ?string
     {
         return preg_replace_callback(
             '/(?:[^' . self::CHAR_UNRESERVED . self::CHAR_SUB_DELIMS . '%:@\/\?]+|%(?![A-Fa-f0-9]{2}))/u',
--- a/test/UriTest.php
+++ b/test/UriTest.php
@@ -634,7 +634,7 @@ class UriTest extends TestCase
     {
         return [
             ['http://example.com/?q=тестовый_путь', 'q=тестовый_путь'],
-            ['http://example.com/?q=ουτοπία', 'q=ουτοπία'],
+            ['http://example.com/?q=ουτοπία', 'q='],
         ];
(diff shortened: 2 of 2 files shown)


src/Uri.php
```diff
@@ -652,7 +652,7 @@ private function filterFragment(string $fragment) : string
     /**
      * Filter a query string key or value, or a fragment.
      */
-    private function filterQueryOrFragment(string $value) : string
+    private function filterQueryOrFragment(string $value) : ?string
```
Ocramius: This kind of change requires a backing test case.

test/UriTest.php
```diff
@@ -634,7 +634,7 @@ public function utf8QueryStringsDataProvider()
     {
         return [
             ['http://example.com/?q=тестовый_путь', 'q=тестовый_путь'],
-            ['http://example.com/?q=ουτοπία', 'q=ουτοπία'],
+            ['http://example.com/?q=ουτοπία', 'q='],
```
Ocramius: This change seems incorrect: why is the data being dropped?

src/Uri.php
soru1001: ok, I've fix a test case.

Ocramius: We'd need an *additional* test case, not a change in existing ones

test/UriTest.php
soru1001: Because, once I change the return type declarations( string -> ?string ), test become a failure.

``` 
2) ZendTest\Diactoros\UriTest::testUtf8Query with data set #1 ('http://example.com/?q=ουτοπία', 'q=ουτοπία')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'q=ουτοπία'
+'q='
```

Ocramius: Yes, which is indicating that the patch caused a regression/BC break, and there is something wrong with it.

webimpress:
@Ocramius @soru1001 

1. I think it is invalid. `preg_replace_callback` returns null on error only: https://www.php.net/preg_replace_callback

2. Even if it would be valid it's a BC Break...

PR #363, opened 2019-07-09
Modify detection of HTTPS

heiglandreas:
This modifies the detection of HTTPS by checking whether the lower-case value of the HTTPS-key of the server variable equals 'on'. Before that checked whether the value did *not* match 'off' which meant that the nginx-default for non-HTTPS connections (which, according to the [documentation](http://nginx.org/en/docs/http/ngx_http_core_module.html#var_https) is an empty string) did in fact match and returned that the connection *is* encrypted.

This fixes #362

Provide a narrative description of what you are trying to accomplish: See #362 

- [x] Are you fixing a bug?
  - [x] Detail how the bug is invoked currently.
  - [x] Detail the original, incorrect behavior.
  - [x] Detail the new, expected behavior.
  - [x] Base your feature on the `master` branch, and submit against that branch.
  - [x] Add a regression test that demonstrates the bug, and proves the fix.
  - [x] Add a `CHANGELOG.md` entry for the fix.

PR #364, opened 2019-07-11
Apply #363 (detection of HTTPS) on v1 (LTS)

webimpress:
As zend-diactoros v1 is still LTS we should provide the hotfix also in that branch.
I've ported @heiglandreas commits with changes to work on v1.

- [x] Are you fixing a bug?

/cc @heiglandreas @weierophinney


Xerkus:
@webimpress, @heiglandreas thank you

PR #366, opened 2019-07-27
Fixed failing tests: set user agent on load external content

webimpress:
For a while tests were failing on the build. Here is the simple fix


webimpress:
Tests are now passing, the only problem is CS issue, introduced in #363 and fixed in #364 which is not merged yet.

PR #367, opened 2019-08-06
Refactor fetching iana http status codes to use caching

Xerkus:
Refactored iana status codes test provider to use locally cached file instead of downloading them on each tests run.

Also fixed failing tests, replaces #366


weierophinney:
I've rebased your commit off of current master, and will merge if the tests pass. Thanks, @Xerkus !

PR #368, opened 2019-09-01
use shorter expressions in functions

pine3ree:
min requirement is now php-7.1 => replace long expressions with equivalent shorter ones (and faster when `array_key_exists` is involved):

- REPLACE `isset($arr[$key]) ? $arr[$key] : $default` WITH `$arr[$key] ?? $default` 
- REPLACE `array_key_exists($key, $arr) ? $arr[$key] : null` WITH `$arr[$key] ?? null`

if an array_key_exists but the associated value is NULL and we use NULL when that key does not exists, the result is equivalent to just calling `isset` and use NULL if not isset, For what is worth...micro-opt...`isset` being as a language construct is faster than the `array_key_exist` func.

The case for:
```php
$unencodedUrl    = $server['UNENCODED_URL'] ?? '';
```
later `! empty($unencodedUrl)` is called so `null` and the empty string `''` are both falsy values anyway and lead to the same result.

### DIFF
--- a/src/functions/create_uploaded_file.php
+++ b/src/functions/create_uploaded_file.php
@@ -33,7 +33,7 @@ function createUploadedFile(array $spec) : UploadedFile
         $spec['tmp_name'],
         $spec['size'],
         $spec['error'],
-        isset($spec['name']) ? $spec['name'] : null,
-        isset($spec['type']) ? $spec['type'] : null
+        $spec['name'] ?? null,
+        $spec['type'] ?? null
     );
 }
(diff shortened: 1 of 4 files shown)


pine3ree:
In my travis build I had a warning unrelated to my changes for non-latest deps, so I guess i will have it here as well:
```
There was 1 warning:
1) Warning
The data provider specified for ZendTest\Diactoros\ResponseTest::testReasonPhraseDefaultsAgainstIana is invalid.
Unable to retrieve IANA response status codes due to timeout or invalid XML
WARNINGS!
Tests: 1019, Assertions: 1727, Warnings: 1.
```

webimpress:
Thanks, @pine3ree!

PR #370, opened 2019-10-08
Hotfix: marshalHeadersFromSapi function

### LINKED ISSUE #369: BUG: marshalHeadersFromSapi assume all SERVER keys are strings
- [x] I was not able to find an [open](https://github.com/zendframework/zend-diactoros/issues?q=is%3Aopen) or [closed](https://github.com/zendframework/zend-diactoros/issues?q=is%3Aclosed) issue matching what I'm seeing.
 - [X] This is not a question. (Questions should be asked on [chat](https://zendframework.slack.com/) ([Signup here](https://zendframework-slack.herokuapp.com/)) or our [forums](https://discourse.zendframework.com/c/questions/expressive).)

For some estrange reason $_SERVER super-global could be populated with an integer key, making `marshalHeadersFromSapi` function to crash.

According to [php manual](https://www.php.net/manual/en/language.types.array.php) integer keys are valid keys:
> The key can either be an integer or a string. The value can be of any type. 

### Code to reproduce the issue

This happened on a Systemd + Nginx + php-pfm, trying to use sentry-symfony 3.2

To reproduce, add this line to the php-fpm pool config:
```
env[42] = testing
```

### Expected results

To deal graciously with integer keys and not crash the request.

### Actual results

```
Fatal error: Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Type error: strpos() expects parameter 1 to be string, int given in /opt/app/vendor/zendframework/zend-diactoros/src/functions/marshal_headers_from_sapi.php:28 Stack trace: 
#0 /opt/app/vendor/zendframework/zend-diactoros/src/functions/marshal_headers_from_sapi.php(28): strpos(92, 'REDIRECT_') 
#1 /opt/app/vendor/zendframework/zend-diactoros/src/ServerRequestFactory.php(64): Zend\Diactoros\marshalHeadersFromSapi(Array) 
#2 /opt/app/vendor/sentry/sentry/src/Integration/RequestIntegration.php(101): Zend\Diactoros\ServerRequestFactory::fromGlobals() 
#3 /opt/app/vendor/sentry/sentry/src/Integration/RequestIntegration.php(74): Sentry\Integration\RequestIntegration->processEvent(Object(Sentry\Event), Object(Sentry\Options)) 
#4 /opt/app/vendor/ in /opt/app/vendor/zendframework/zend-diactoros/src/functions/marshal_headers_from_sapi.php on line 28
```

webimpress:
- integer keys should be ignored
- _ should be changed to - in key (also for CONTENT_* headers)

Fixes #369 

- [x] Are you fixing a bug?
  - [ ] Add a `CHANGELOG.md` entry for the fix.

### DIFF
--- a/src/functions/marshal_headers_from_sapi.php
+++ b/src/functions/marshal_headers_from_sapi.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace Zend\Diactoros;
 
 use function array_key_exists;
+use function is_string;
 use function strpos;
 use function strtolower;
 use function strtr;
@@ -23,6 +24,10 @@ function marshalHeadersFromSapi(array $server) : array
 {
     $headers = [];
     foreach ($server as $key => $value) {
+        if (! is_string($key)) {
+            continue;
+        }
+
         // Apache prefixes environment variables with REDIRECT_
         // if they are added by rewrite rules
         if (strpos($key, 'REDIRECT_') === 0) {
--- a/src/functions/marshal_headers_from_sapi.php
+++ b/src/functions/marshal_headers_from_sapi.php
@@ -45,7 +45,7 @@ function marshalHeadersFromSapi(array $server) : array
         }
 
         if (strpos($key, 'HTTP_') === 0) {
-            $name = strtr(strtolower(substr($key, 5)), '_', '-');
+            $name = str_replace('_', '-', strtolower(substr($key, 5)));
             $headers[$name] = $value;
             continue;
         }
--- a/src/functions/marshal_headers_from_sapi.php
+++ b/src/functions/marshal_headers_from_sapi.php
@@ -28,6 +28,10 @@ function marshalHeadersFromSapi(array $server) : array
             continue;
         }
 
+        if ($value === '') {
+            continue;
+        }
+
         // Apache prefixes environment variables with REDIRECT_
         // if they are added by rewrite rules
         if (strpos($key, 'REDIRECT_') === 0) {
@@ -38,10 +42,6 @@ function marshalHeadersFromSapi(array $server) : array
             if (array_key_exists($key, $server)) {
                 continue;
             }
-        }
-
-        if ($value === '') {
-            continue;
         }
 
         if (strpos($key, 'HTTP_') === 0) {
--- a/src/functions/marshal_headers_from_sapi.php
+++ b/src/functions/marshal_headers_from_sapi.php
@@ -45,13 +45,13 @@ function marshalHeadersFromSapi(array $server) : array
         }
 
         if (strpos($key, 'HTTP_') === 0) {
-            $name = str_replace('_', '-', strtolower(substr($key, 5)));
+            $name = strtr(strtolower(substr($key, 5)), '_', '-');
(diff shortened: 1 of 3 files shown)


src/functions/marshal_headers_from_sapi.php
```diff
@@ -35,18 +44,14 @@ function marshalHeadersFromSapi(array $server) : array
             }
         }
 
-        if ($value === '') {
-            continue;
-        }
-
         if (strpos($key, 'HTTP_') === 0) {
-            $name = strtr(strtolower(substr($key, 5)), '_', '-');
+            $name = str_replace('_', '-', strtolower(substr($key, 5)));
```
weierophinney: Why have you changed this? It's unrelated...

src/functions/marshal_headers_from_sapi.php
```diff
@@ -35,18 +44,14 @@ function marshalHeadersFromSapi(array $server) : array
...
             $headers[$name] = $value;
             continue;
         }
 
         if (strpos($key, 'CONTENT_') === 0) {
-            $name = 'content-' . strtolower(substr($key, 8));
+            $name = str_replace('_', '-', strtolower($key));
```
weierophinney: Why have you changed this? It's unrelated...

src/functions/marshal_headers_from_sapi.php
webimpress: there is no any functional change, unrelated, but just to be inline with the other change below

src/functions/marshal_headers_from_sapi.php
webimpress: See commits, I've written test first for the whole behaviour of the method I would expect, and then I've made a changes to fulfil test. Yes, it seems to be separate issue.

webimpress: When we have multiple `_` in the key - it is possible for `HTTP_*` headers, but not sure now if for `CONTENT_*`. I know Content-Security-Policy or Content-Transfer-Encoding, but these are response headers, not request. So maybe it is not a case?

weierophinney: Hmm, that's interesting. My understanding was that `strtr()` would replace all instances - and a quick test of the following shows it definitely does:

```php
$string = 'CONTENT_SECURITY_POLICY';
$revised = strtr('CONTENT_SECURITY_POLICY', '_', '-'); // 'CONTENT-SECURITY-POLICY'
```

(see https://3v4l.org/uEKEQ for an example; I also do `strtolower()` in that one)

weierophinney: Perhaps it's the `substr()` call we need to remove instead?

webimpress: @weierophinney there was no strtr in that line ...
I've removed concatenation and substr and added str_replace

weierophinney: So, since it's new, let's go with `strtr()`, as that would make it consistent with the previous usage.

webimpress: Updated

PR #372, opened 2019-10-09
Fix #357 - invalid characters in path/query

### LINKED ISSUE #218: Character encoding issue in Uri.php filterQuery() method.

### LINKED ISSUE #357: Uncaught TypeError: Return value of Zend\\Diactoros\\Uri::filterQueryOrFragment() must be of the type string, null returned

webimpress:
Fixes #357 
Fixes #218

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -11,9 +11,7 @@ namespace Zend\Diactoros;
 
 use Psr\Http\Message\UriInterface;
 
-use function array_key_exists;
 use function array_keys;
-use function count;
 use function explode;
 use function get_class;
 use function gettype;
@@ -23,10 +21,12 @@ use function is_object;
 use function is_string;
 use function ltrim;
 use function parse_url;
+use function preg_match;
 use function preg_replace;
 use function preg_replace_callback;
 use function rawurlencode;
 use function sprintf;
+use function str_split;
 use function strpos;
 use function strtolower;
 use function substr;
@@ -560,6 +560,8 @@ class Uri implements UriInterface
      */
     private function filterUserInfoPart(string $part) : string
     {
+        $part = $this->filterInvalidUtf8($part);
+
         // Note the addition of `%` to initial charset; this allows `|` portion
         // to match and thus prevent double-encoding.
         return preg_replace_callback(
@@ -574,6 +576,8 @@ class Uri implements UriInterface
      */
     private function filterPath(string $path) : string
     {
+        $path = $this->filterInvalidUtf8($path);
+
         $path = preg_replace_callback(
             '/(?:[^' . self::CHAR_UNRESERVED . ')(:@&=\+\$,\/;%]+|%(?![A-Fa-f0-9]{2}))/u',
             [$this, 'urlEncodeChar'],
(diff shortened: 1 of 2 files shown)


webimpress:
@weierophinney I was looking for right solution of this issue, and I think I have found one finally.
In my opinion we need to do something with invalid characters - in that case encode - what I mean we can't just throw exception, as these can be user input.
So if invalid UTF-8 character is detected in the string, then I am checking the string letter by letter to encode that invalid character. This solution run regexp on one character in the loop, but I couldn't find anything better. Any suggestions appreciated. 

/cc @YellowPepper @krowinski @freyr @oroszlanyzsolt

src/Uri.php
```diff
@@ -594,6 +598,25 @@ private function filterPath(string $path) : string
...
+    /**
+     * Encode invalid UTF-8 characters in given string. All other characters are unchanged.
+     */
+    private function filterInvalidUtf8(string $string) : string
+    {
+        if (preg_match('//u', $string)) {
+            return $string;
+        }
+
+        $letters = str_split($string);
```
weierophinney: Noting here, because I had to look it up - `str_split()` splits into bytes instead of characters when it encounters a multibyte string.

src/Uri.php
```diff
@@ -594,6 +598,25 @@ private function filterPath(string $path) : string
         return '/' . ltrim($path, '/');
     }
 
+    /**
+     * Encode invalid UTF-8 characters in given string. All other characters are unchanged.
+     */
+    private function filterInvalidUtf8(string $string) : string
+    {
+        if (preg_match('//u', $string)) {
```
weierophinney: I don't understand what this is doing, exactly.

I've tried it locally with a variety of strings from your query provider below, and every single one of them results in a positive match, which means the logic below never gets hit. Maybe you're not testing _invalid_ UTF-8 characters anywhere?

webimpress: How come?

See: https://3v4l.org/MEYgF

```php
var_dump(preg_match('//u', "\x21\x92"));
```

result:
```
bool(false)
```

Somehow test I've added works.

weierophinney: Turned out I was using single quotes instead of double quotes when I was doing my tests here. Once I changed the quoting style, I was able to observe the same behaviour finally.

Might be good to drop a note in indicating what this is doing, though, so future contributors/maintainers know the purpose.

krowinski:
tx for fix!

PR #376, opened 2019-11-05
`X-Forwarded-Host` for `marshalUriFromSapi`

boesing:
Hey guys,

we are using NGINX+ with loadbalancing to multiple servers which are not aware of external domain names (for reasons).
However, NGINX provides `X-Forwarded-Host` header to those upstream servers to tell them which external domain was called.

I've added an example on what I would like to expect (even tho that `X-Forwarded-Proto` is being handled for the `https` scheme).

Tell me what you think.

### DIFF
--- a/src/functions/marshal_uri_from_sapi.php
+++ b/src/functions/marshal_uri_from_sapi.php
@@ -80,7 +80,7 @@ function marshalUriFromSapi(array $server, array $headers) : Uri
         * @return array Array of two items, host and port, in that order (can be
         *     passed to a list() operation).
         */
-        $marshalIpv6HostAndPort = function (array $server, string $host, ?int $port) : array {
+        $marshalIpv6HostAndPort = function (array $server, ?int $port) : array {
             $host = '[' . $server['SERVER_ADDR'] . ']';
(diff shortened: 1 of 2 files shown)


michalbundyra:
@boesing it makes sense to me.

here is very old PR to fix it in ZendView helper: https://github.com/zendframework/zendframework/pull/2279/files

boesing:
I've implemented the way on how that might work.
Just a sidenote: I've skipped that "explode" stuff as per [documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) this header only contains a single host (and thats, in fact, is the case for NGINX+ tho).
Probably, the guy who implemented that in  https://github.com/zendframework/zendframework/pull/2279 was confused by the [X-Forwarded-For](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) header.

boesing:
Would appreciate some review.

/cc @michalbundyra @froschdesign

weierophinney:
Thanks, @boesing! Will release once #378 is complete.

Issue #377, opened 2019-11-08
Can UploadedFile extend SplFileInfo ?


mattsah:
Is there a particular reason that UploadedFile does not extend SplFileInfo.  The only conflicting interface, which isn't even really a conflict so much is `getSize()` -- would there be any interest in this?  I can do a PR if the interest is there.

I need this feature like yesterday and without it I'm either gonna need to look for an alternative or fork.

weierophinney:
@mattsah Send a pull request, and we can evaluate!

mattsah:
#378

froschdesign:
Another notice from @michalbundyra: (https://3v4l.org/DnSXf)

```php
class A {
    public function getSize(): int {
        return 2;
    }
}
class B extends A {
    public function getSize(): ?int
    {
        return null;
    }
}

$b = new B;
var_dump($b->getSize());
```

This results in:

> Fatal error: Declaration of B::getSize(): ?int must be compatible with A::getSize(): in

---

At the moment `SplFileInfo` does not define an explicit return type for `getSize` method but it could become a problem in the future.

mattsah:
@froschdesign 

I'm not sure I understand the first point regarding StreamInterface.  That is, I don't see the relevance with respect to UploadedFile.  UploadedFile appears to return ?int (from your own link), null if unknown.  If a stream is passed to UploadedFile, even if it called getSize() directly on the stream, if the return typeof that is always ?int, then it falls within the appropriate return type for UplaodedFile.

Regarding SplFileInfo.  Per PSR-7 spec, the getSize() is intended to return the size as actually transferred:

```
* Implementations SHOULD return the value stored in the "size" key of
* the file in the $_FILES array if available, as PHP calculates this based
* on the actual size transmitted.
```

Assuming this is accurate, the size transferred, should always be 0 or > 0 -- which means, while UploadedFile may not throw an exception the same as SplFileInfo, it will always report the actual size of an uploaded file.

I think the case I'm not understanding is where UploadedFile::getSize() would return NULL -- this would certainly conflict with SplFileInfo.  But under what condition does this occur?

In either case, neither UploadedFile or SplFileInfo _actually_ typehint return.  If PHP is willing to break backwards compatibility, I'd think all bets are off for libraries and you're in a position to address BC breaks regardless.  If PSR-7 changes the actual programmatic interface, presumably that would require a potentially significant update anyway, including requiring newer versions of PHP.

ADmad:
[...]

PR #378, opened 2019-11-08
Make UploadedFile Extend SplFileInfo

mattsah:
This is a new feature to make UploadedFile extend SplFileInfo which enables you to pass UploadedFile instances in places where SplFileInfo typehints are required and use SplFileInfo based methods to access information about the file while still having access to the additional UploadedFile interface.

This is primarily useful for populating models/entities directly from requests wherein a supplied value needs to be an instance of SplFileInfo and wherein a service or hook later may move the files into more appropriate places at time of persistence.

### DIFF
--- a/src/UploadedFile.php
+++ b/src/UploadedFile.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 namespace Zend\Diactoros;
 
+use SplFileInfo;
 use Psr\Http\Message\StreamInterface;
 use Psr\Http\Message\UploadedFileInterface;
 
@@ -35,7 +36,7 @@ use const UPLOAD_ERR_NO_TMP_DIR;
 use const UPLOAD_ERR_OK;
 use const UPLOAD_ERR_PARTIAL;
 
-class UploadedFile implements UploadedFileInterface
+class UploadedFile extends SplFileInfo implements UploadedFileInterface
 {
     const ERROR_MESSAGES = [
         UPLOAD_ERR_OK         => 'There is no error, the file uploaded with success',
@@ -102,9 +103,13 @@ class UploadedFile implements UploadedFileInterface
         if ($errorStatus === UPLOAD_ERR_OK) {
             if (is_string($streamOrFile)) {
                 $this->file = $streamOrFile;
+
+                parent::__construct($this->file);
             }
             if (is_resource($streamOrFile)) {
                 $this->stream = new Stream($streamOrFile);
+
+                parent::__construct($this->stream->getMetaData('uri'));
             }
 
             if (! $this->file && ! $this->stream) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,33 @@
 
 All notable changes to this project will be documented in this file, in reverse chronological order by release.
 
-## 2.2.0 - TBD
+## 2.2.1 - TBD
 
 ### Added
 
 - [#376](https://github.com/zendframework/zend-diactoros/pull/376) adds support for using the X-Forwarded-Host header for determining the originally requested host name when marshaling the server request.
+
+### Changed
+
+- Nothing.
+
+### Deprecated
+
+- Nothing.
+
+### Removed
+
+- Nothing.
+
+### Fixed
+
+- Nothing.
+
+## 2.2.0  - 2019-11-08
+
+### Added
+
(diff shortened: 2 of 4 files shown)


michalbundyra:
@mattsah we need to add some unit tests for these changes

mattsah:
Could you expand on what sorts of tests might be meaningful here?  Happy to add them, but I'm a bit at a loss of what precisely to unit test as SplFileInfo is part of PHP itself.  I don't know that I fully understand what case needs to be covered, but I may be less familiar with all the ways in which UploadedFile is used.

weierophinney:
> Could you expand on what sorts of tests might be meaningful here? Happy to add them, but I'm a bit at a loss of what precisely to unit test as SplFileInfo is part of PHP itself.

In this case, a test that verifies that an `UploadedFile` is also an instance of `SplFileInfo`. Having the test case will ensure that we do not remove that inheritance in a later commit.

weierophinney:
@mattsah I've pushed a change containing a test for you. :smile:

michalbundyra:
@weierophinney I would expect more tests here, as we call parent constructor with different values depends on some conditions. And there is a case when we do not call the parent constructor. I believe all these should be covered in tests.

src/UploadedFile.php
```diff
@@ -102,9 +103,13 @@ public function __construct(
         if ($errorStatus === UPLOAD_ERR_OK) {
             if (is_string($streamOrFile)) {
                 $this->file = $streamOrFile;
+
+                parent::__construct($this->file);
             }
             if (is_resource($streamOrFile)) {
                 $this->stream = new Stream($streamOrFile);
+
+                parent::__construct($this->stream->getMetaData('uri'));
```
ADmad: Does `SplFileInfo` works with stream paths? PHP docs say the argument for `SplFileInfo::__construct()` should be a file path while the value of `$this->stream->getMetaData('uri')` could be a stream path `'php://memory'`.

As commented by @michalbundyra this patch should have more tests.

ADmad: As suspected things go kaboom when trying to use some of the `SplFileInfo` methods if it's created with a stream path.

https://3v4l.org/klK0X

So I am not sure if this is a very good change.

froschdesign: @ADmad
Thanks for your comment! 👍 

We have another discussion in the related issue report: #377

mattsah: @ADmad nowo run getMTime() on a file that doesn't exist... SplFileInfo is doing exactly what it is supposed to do here.  This is not going "kaboom" -- this is how it works.  It's not intended to provide information about files that it can't get information for -- which includes any non existent file.  Your test seems to be based on a false assumption about SplFileInfo that it requires an actually existing file... it doesn't.  SplFileInfo is designed to grok information about files, which means the path you pass it MAY not exist, and it's up to a developer to use it properly.

There may, of course be edge cases that we'd want to handle differently.  But the example provided is just SplFileInfo working as it does.

ADmad: > SplFileInfo is designed to grok information about files

Exactly, "about files" not streams.

weierophinney: @mattsah —

Having the class extend `SplFileInfo` implies that the item returned has a relationship with the filesystem, and can be operated as if it were part of the filesystem. It also implies that users should be able to safely call its various methods.

Considering the major use case is to use a temporary stream (Diactoros creates temporary streams by default), we absolutely need stream support to work without errors. Since it wasn't, we had to revert.

mattsah: @weierophinney 

Is there a single case where SplFileInfo acts differently on a stream than it does on a non-existent file?  It seems quite strange to me that Spl**File**Info suggests a relationship with the filesystem while Uploaded**File** does not.  A stream has the same relationship to the filesystem as a non-existent file.  Again, the implication here seems to be that SplFile**Info** is not about grokking information.

Additionally, if one is truly concerned with supporting streams more fully, it would be perfectly fine, for example, to overload getMTime() later and call parent if its a file or examine stream meta data if it's a stream.

PHP uses stream URLs in places of "files" in all sorts of cases... the implication you suggest doesn't seem well supported.

Regarding major use cases and defaults... given that this did not previously extend it, I don't see any case where anyone would be relying on SplFileInfo methods... stream support would absolutely continue working as it has historically and **could be improved upon** to offer even more info via overloads as mentioned above in the future.

This is non-breaking, working as expected, and providing a useful feature.  It may not provide all the features you want out of the gate, i.e. functional mtime on stream, but that'd make for a nice 2.2.1 -- awell.

Very disappointed here.

